PR #1, opened 2023-07-05
Api history

deitch:
Moves api out of `lf-edge/eve/api/` and into here at the root. Moves the entire history over (for commits that affect this directory) as well.


uncleDecart:
I think it's great that we will have separate API repo, thank you for your efforts @deitch!
But that actually creates questions which are in my opinion are crucial

- Do we want to finally create versioning for EVE API? And if yes, should we write somewhere visible enough how we are doing it (important things like backward compatibility, minor, major versions, is EVE version is the same as EVE API etc.)?
- Do we want to formalize, when we are switching to new version in EVE API or is it just follow master?

They are crucial because they will define our workflow when you have to change API :)

deitch:
These are very good questions. I don't have an immediate opinion on them. I do think the first thing we need to do is get them moved over here, so it looks exactly like it did in the single eve repo. That requires a window when the api is not changing. 

After that, we can start a separate strategy for api.

deitch:
@eriknordmark do you have views?

eriknordmark:
> I think it's great that we will have separate API repo, thank you for your efforts @deitch! But that actually creates questions which are in my opinion are crucial
> 
> * Do we want to finally create versioning for EVE API? And if yes, should we write somewhere visible enough how we are doing it (important things like backward compatibility, minor, major versions, is EVE version is the same as EVE API etc.)?
> * Do we want to formalize, when we are switching to new version in EVE API or is it just follow master?
@uncleDecart 
We do have a versioning approach (we are currently at v2) but we don't have any of that documented.
If someone wants to take the pen and start on versioning section in api/README.md?
It would make sense to cover incompatible changes (such as those which made us move from v1 to v2), additions (which we do all the time without any explicit versioning) and additions to the config API which the controller needs to know about (which is where we use the api_capabilities field in the info API).

I think de-facto the EVE API is v2 hence different that the EVE-OS version, but it is a good question whether we should add some explcit minor or micro version to the EVE API which is increamented on compatible changes and have EVE-OS report that e.g., in the hw info message sent on boot.

deitch:
The DCO will not be happy because of some of the much older commits. It will require an admin to override it @eriknordmark

eriknordmark:
@deitch I think here the assumption is that the developer does a make, and then commits all of the generated files.
However, the README doesn't specify what the developer is supposed to do.

And I think it makes sense to add a GH workflow which runs on PRs which does a make and then makes sure that no files change or are added as part of the make i.e., all of the derived files are checked in. 

Alternatively we do check in the derived files (and have the users of the repo do that) in which case we should check that proto is empty.

uncleDecart:
Hey @deitch do you think it makes sense to create separate `doc` folder to store all documentation regarding API or should we just link documents from EVE repository?

deitch:
I think you are right; we should move the docs over. But this was hard enough to get right, so I want this in, and then I can do a separate single-commit PR for docs. 

It's in the hands of @eriknordmark now.

eriknordmark (approved):
LGTM.

But we also need to get the workflows in place. I'll work on that later this week but can you do a separate PR to add the yetus files from lf-edge/eve so we can do the same runs?
Then I can figure out how to add the GH actions to run yetus and make all tests.

eriknordmark (approved):
I see the workflow files are in place.

eriknordmark (commented):
@deitch looks like the workflow failed in
https://github.com/lf-edge/eve-api/actions/runs/5638807647/job/15273322681#step:5:8

deitch:
Urgh, that is my mistake. I will open a PR to clean that up.

deitch:
@eriknordmark see #2

PR #2, opened 2023-07-24
use correct remote package name in actions

deitch:
fix broken `go install` in actions


deitch:
I have the various install dependencies. It does check that nothing has changed, i.e. that all the various generations were done. That means going through the generated files that have changed, to ensure it all is compatible.

deitch:
Ran against a consistent v3.21.9, which was the last run on this prior to switch. It _should_ pass now.

deitch:
And now it appears happy.

eriknordmark (approved):
LGTM

PR #3, opened 2023-07-24
rename path from eve/api to eve-api

deitch:
It still used the old paths, which made the import fail.


deitch:
I suspect the CI may not clear until #2 is in.

deitch:
Rebased.

PR #5, opened 2023-07-26
add CODEOWNERS

deitch:


eriknordmark (approved):
LGTM, but we should find and add additional reviewers to this later.

deitch:
Agreed. I just took what was there for now. I do not know who the right CODEOWNERS should be.

PR #6, opened 2023-07-26
Add format type PVC

zedi-pramodh:
With kubevirt eve the storage format for volumes will be of type PVC. So add PVC format to storage.proto

Rest of the files are coming from "make proto"


zedi-pramodh:
storage.proto is the only file that changed. Rest all are generated from "make proto" 
Sure I can split to different commits if it helps reviewing.

zedi-pramodh:
Split the commit as per review comments

uncleDecart (approved):
LGTM
Thank you for splitting commits @zedi-pramodh ! I wonder if we should specify that PVC is used for Kubernetes or is it common knowledge

eriknordmark (approved):
LGTM but conflicts in the generated files.

zedi-pramodh:
I have no idea on how these conflicts came and how can I resolve those. I just ran "make proto"

zedi-pramodh:
Looks like for every comment line an extra space got added after //

eriknordmark:
> I have no idea on how these conflicts came and how can I resolve those. I just ran "make proto"

Some other PR updated an unrelated proto file.
This is why a separate commit with the generated files makes sense.

This is how I normally resolve these:
1. git rebase -i upstream/master; mark your second commit with a "drop" (or go a git reset to drop it)
2. git fetch upstream master
3. git rebase -i upstream/master; should have no conflicts since you only have one added line in the proto file
4. make proto; git commit -s, etc

zedi-pramodh:
Updated the commit after rebase to upstream/main.

PR #8, opened 2023-07-27
fix bash error in CI

deitch:
Note that this may trigger CI failures because of regenerated files, but that would mean this is fixed, and should go in.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -41,7 +41,7 @@ jobs:
           # we run it twice, so that it will report on the logs
           echo "changed and uncommitted files"
           git status --short          
-          if [ $(git status --short) != '' ]; then
+          if [ -n "$(git status --short)" ]; then
             # print the diff so we can fix it, if there is an issue
             git diff
             exit 1


deitch:
The test itself actually works now, and it triggers CI failures. It does so because CI runs `make proto`  to check that the generated language bindings and graphviz are up to date. This fails, even though they are up to date, for 2 reasons:

* The graphviz changes with each run. I am not sure why, we will need to figure that out.
* If your version of `protoc` and `protoc-gen-go` are different than what CI runs, it will modify the files

Going to need to find ways around these.

For the second issue, e.g.:

```go
--- a/go/attest/attest.pb.go
+++ b/go/attest/attest.pb.go
@@ -3,8 +3,8 @@
 
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
-// 	protoc-gen-go v1.28.1
-// 	protoc        v4.23.4
+// 	protoc-gen-go v1.31.0
+// 	protoc        v3.21.9
 // source: attest/attest.proto
 
 package attest
``

I ran with different versions of the tools than CI, which gets embedded in and causes changes. I can see a few possibilities:

1. Maybe we can filter out just for that? i.e. ignore those changes? It will take some git magic work and some weird filtering.
2. Maybe we can lock in particular versions? Thus, you are required to use that version locally? That isn't great, what if you have one version, I have another, and the CI has a third? We shouldn't be tightly bound to a very specific version.

Looking for ideas.

For the first, I still need to see why it is regenerating.

eriknordmark (approved):
LGTM

eriknordmark:
@deitch can we just ignore the graph in the comparison?
(In addition we need some way to skip the version strings - maybe we should have the action do the make proto and commit the output instead of the user doing that?)

deitch:
> maybe we should have the action do the make proto and commit the output instead of the user doing that?)

I thought about that. I rather like it, but commits from within CI? And if the original creator needs to change something? Ugh.

> can we just ignore the graph in the comparison?

Not really. We want to ensure it was generated.

PR #9, opened 2023-07-28
add CI-based generation and commit of language bindings and visualizations

deitch:
Because the visualizations are not consistent ordering, generating them and checking if they were correct will give an error consistently.

Because the language bindings, specifically go, include the version, even minor changes on the committer side would show as changes.

We solve both of these by having CI generate those bindings - overriding anything the user did - and committing them back.


deitch:
Looks like this won't work. Oh well.

PR #10, opened 2023-07-30
small simplification of protodot generation; reduced error noise in CI

deitch:
Small improvement in how the makefile calls generating the graphviz; no change in output.

More importantly, ignores differences in versions in the comments when checking if the files were generated properly. Of course, some changes in versions can lead to more fundamental changes in the content, so when it finds something other than a version-in-the-comment diff, it reports both diffs: ignoring versions, and entire diff.


deitch:
This should solve most of the extraneous errors in CI, but still doesn't solve the graph issue. Working on that in a future PR.

deitch:
Confirmed. If you check the output of CI on this one, the only diffs left are the generated graph images. I am working on that, will get to it in a separate PR when it is ready. It should not block this one.

eriknordmark (approved):
LGTM

PR #12, opened 2023-08-07
[WIP] Update configuration services

uncleDecart:
This PR contains small changes for introduced patchEnvelope entity.
Note that some changes are breaking compatibility with previous version of API, but since there's nobody using it yet, it's not a problem.

changelog:
- Instead of using Volume configuration we are using volume ref
- Adding per app usage statistics for patch envelopes


uncleDecart:
cc: @gkodali-zededa

gkodali-zededa (approved):

proto/config/patch_envelope.proto
```diff
@@ -35,7 +35,7 @@ message EveOpaqueBase64Data {
 
 message EveOpaqueBinaryBlob {
     // Target field of Volume set to AppCustom
-    Volume opaqueVolume = 1;
+    string opaqueVolume = 1;
```
eriknordmark, line 38: We clearly show enable yetus/protocmp (I think that is what we were running in the EVE repo) to verify that it will both flag this as an incompatible type change (which we can ignore here since as you say this current message in the API has never been used) and also flag that the field names should be in lower_snake_case and not in camelCase. (We should follow that for all new fields.)

Can you or @deitch add yetus to do those checks?

deitch:
Based on [this diff](https://github.com/lf-edge/eve-api/actions/runs/5782883418/job/15670575126?pr=12#step:9:38), I think you are using a different version of `protoc-gen-go`. Can you check the version? Or better yet, run it in the devcontainer?

proto/config/patch_envelope.proto
eriknordmark: @deitch according to https://github.com/lf-edge/eve-api/actions/workflows/yetus.yaml it has never run, and the workflow refers to "master" and not "main". Makes sense to grep for master in both eve-api and eve-libs.

eriknordmark, line 38: @uncleDecart is this reference a UUID? If so it would make sense to state that in a comment.

proto/info/patch_envelope.proto
```diff
@@ -61,3 +61,22 @@ message ZInfoOpaqueAppInstanceStatus {
...
+
+// Statistics on usage of specific patchEnvelope for specific app
+message ZInfoPatchEnvelopeApp {
+    // UUID of PatchEnvelope
+    string uuid = 1;
+    // Version of PatchEnvelope
+    string version = 2;
+    // Number of times this PatchEnvelope
+    // was sucessfully downloaded
+    uint64 timesDownloaded = 3;
```
eriknordmark, line 81: What did we decide regarding updates to a patch envelope? Will the controller update it and use the same UUID? Or generate a new UUID for the updated PE?

The download counter makes some sense in the former case.
Is this the number of times the app instance has retrieved the PE? 

Makes sense adding some more comments to make the intended semantics and usage more clear.

proto/config/patch_envelope.proto
deitch: Good point. I will update the list to master and main

deitch:
This PR is out of date, needs to be rebased.

proto/config/patch_envelope.proto
deitch: And remove those quasi-semver-ish branches, as that is an eve thing, not an eve-api thing

uncleDecart: Yes it is, I'll comment it out

uncleDecart: Thanks @deitch for #21!

uncleDecart:
> Based on [this diff](https://github.com/lf-edge/eve-api/actions/runs/5782883418/job/15670575126?pr=12#step:9:38), I think you are using a different version of protoc-gen-go. Can you check the version? Or better yet, run it in the devcontainer?

@deitch I actually had very interesting experience with devcontainer, when I was rebasing this branch to main I had merge conflicts with generated proto files and I could only remove them with sudo. Is there a better way to do this?

deitch:
Only remove them with `sudo`? That is strange. Are you using `docker run` or the devcontainer cli or vs code or GitHub workspaces?

uncleDecart:
devcontainer cli

deitch:
We probably should explicitly specify the user in the config. I never am sure what it should be, as everyone has a different user on their desktop.

Well, I will do something with it in a PR.

I really like the _idea_ of devcontainers. We do something very similar (in a very complex way) with eve, and I have done it in dozens of other projects. The experience has been a learning process, though.

uncleDecart:
Completely agree with you @deitch , it is very good _concept_ but it takes time to make good implementation of it :)

deitch:
@uncleDecart I kind of am struggling what to do with this. The problem is that we generate artifacts inside the container (which is fine), and are output to a mounted directory. Since the container is run as root, the output artifacts may be owned by root as well.

However, when I ran it, I didn't see any issues. That could be because I am on Mac, and maybe the translation layer handles it?

In terms of git clone, I didn't see any issues at all. Everything is owned by whatever user clones it.

How do we recreate the issue you have?

uncleDecart:
Clone this branch. Build proto with devcontainer, checkout to main, pull latest and then try to rebase this branch on main :)

deitch:
> Clone this branch. Build proto with devcontainer, checkout to main

How? Your branch is far enough behind that it does not even have the devcontainer config. [Here is the branch](https://github.com/uncleDecart/eve-api/tree/update-configuration-services).

```bash
$ git fetch origin pull/12/head:update-configuration-services
From https://github.com/lf-edge/eve-api
 * [new ref]         refs/pull/12/head -> update-configuration-services
$ git co update-configuration-services
Switched to branch 'update-configuration-services'
$ git diff main --name-only
.devcontainer/Dockerfile
.devcontainer/devcontainer.json
.github/workflows/ci.yaml
Makefile
...
```

Nothing to do there.

proto/config/patch_envelope.proto
uncleDecart: Done

proto/info/patch_envelope.proto
uncleDecart: We do the former, that's why it makes sense

proto/info/info.proto
```diff
@@ -917,6 +917,7 @@ message ZInfoApp {
   repeated ZInfoNetwork network = 16;       // up/down; allocated IP
   repeated string volumeRefs = 17;          // volume UUIDs
   repeated ZInfoSnapshot snapshots = 18;    // optional field, used to send list of snapshots on device
+  ZInfoPatchEnvelopeStat patchEnvelope = 19; // optional; information about every patchEnvelope assigned to app
```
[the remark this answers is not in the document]
uncleDecart: Changed name, decided not to move them to ZMetrics message, I believe it makes more sense to clarify it with more comments

uncleDecart:
Rebased and squashed all commits in one. Made changes in API to be compatible with our controller @udit-zededa please check :)

uncleDecart:
@eriknordmark @deitch if you don't have any objections, let's merge this PR

proto/config/patch_envelope.proto
```diff
@@ -9,61 +9,68 @@ option java_package = "org.lfedge.eve.config";
...
+    // the same file name shall be advertised to application agent
+    // if it's empty EVE will use image name
+    optional string fileNameToUse = 2;
+
+    // can be image type or size encoded into a single string
+    optional string blobMetaData = 3;
+
     // Target field of Volume set to AppCustom
-    Volume opaqueVolume = 1;
+    // reffers to UUID of a volume
```
eriknordmark, line 44: Typo reffers. Do you want to fix before I merge?

eriknordmark (approved):
LGTM but found one typo in a comment.

uncleDecart:
Let's merge it. I'll fix comment later

Issue #11, opened 2023-08-03
Makefile doesn't work manually


eriknordmark:
Back in eve/api I could do a make proto and it ran protoc etc.
That was using the go builder.

When I try that (under Linux) in the eve-api repo it fails with
nordmark@bobo:~/main/eve-api$ make proto
rm -rf go/*/; mkdir -p go # building proto-api-go
protoc -I./proto --go_out=paths=source_relative:./go proto/*/*.proto
/bin/sh: 1: protoc: not found
make: *** [Makefile:19: proto-api-go] Error 127
nordmark@bobo:~/main/eve-api$ which protoc
nordmark@bobo:~/main/eve-api$ make
protodot -inc /usr/include -src ./proto/config/devconfig.proto -output devconfig -generated ./images
make: protodot: No such file or directory
make: *** [Makefile:2: proto-diagram] Error 127
nordmark@bobo:~/main/eve-api$ 

Can we make this work, or add the prerequisite installs to the README.md?

deitch:
The prerequisites are listed in the README [here](https://github.com/lf-edge/eve-api#prerequisites).

The other question is whether we want to containerize this so that it is all set ups, the same way we did in eve? I don't think we want that heavy go-builder for this, we could do something much more lightweight.

Is it worth it?

eriknordmark:
On 8/6/23 02:57, Avi Deitcher wrote:
> The prerequisites are listed in the README here 
> <https://github.com/lf-edge/eve-api#prerequisites>.

Either we introduced a bug recently, or those instructions are insufficient.
After installing them I get
***@***.***:~/main/eve-api$ make proto
rm -rf go/*/; mkdir -p go # building proto-api-go
protoc -I./proto --go_out=paths=source_relative:./go proto/*/*.proto
config/patch_envelope.proto: This file contains proto3 optional fields, 
but --experimental_allow_proto3_optional was not set.
make: *** [Makefile:19: proto-api-go] Error 1

Are we running "make" or "make proto" as part of the CI?

   Erik

deitch:
Then incomplete. 

Can you post here what versions of the tools you have?

Yes CI is running the tools as well. I suspect minimum version or flags or such. Let's use this failure to narrow down the issue. 

Maybe we should make it run-in-docker with an option to run locally? Rather than the current run locally? I'll look into that. 

Either way, do post your versions.

eriknordmark:
> Either way, do post your versions.

FWIW
protoc --version
[...]

PR #13, opened 2023-08-07
buildable in container

deitch:
Adds devcontainer support to eve-api, so it can be run with all tools ready and installed. Of course, it is not required. You always can run `make proto` locally, but this gives a standard way to run it.

Updates `README.md` as well.


deitch:
Fixes #11

deitch:
The only error in CI now is the known issue with the generated graph visual being different.

eriknordmark (approved):
LGTM

PR #14, opened 2023-08-16
Encrypted AuthContainer payload

rouming:
AuthContainer is the envelope for all messages coming from the controller or sent by EVE. For the LOC case the node config should be encrypted. The best place to provide encryption is to modify the `AuthContainer` itself, so that the message can be both signed and encrypted (in this exact order).
    
This patch adds the `cipherData` and `cipherContext` members to the `AuthContainer` to encrypt the payload. The decrypt and validate procedure is the following:
    
1. if `cipherData` is valid the encrypted payload is defined in the `cipherData.cipherData` member, the `protectedPayload` is set to null.
1. Once EVE stack decrypts the payload, `protectedPayload` member is set with the decrypted data.
1. The signature verification procedure can follow.


proto/auth/auth.proto
```diff
@@ -18,9 +19,14 @@ message AuthContainer {
   AuthBody protectedPayload = 1;      // envelope body, a marshalled protobuf data or it can be null
   // if the length of senderCertHash received is not N bytes, as described in hashAlgorithm, then the protobuf
   // message either is not AuthContainer type, or is corrupted. Otherwise, the
-  // receiver may not have the sender's signing certificate
+  // receiver may not have the sender's signing certificate.
+  // If `cipherData` is defined and valid, then `protectedPayload` is null and actual payload is
+  // defined in the `cipherData.cipherData` member. Once the `CipherBlock` is decrypted,
+  // `protectedPayload` is defined by the auth stack.
```
eriknordmark, line 25: I know what you're trying to say but I don't know who will understand it. Would it make sense to re-word to say e..g., Once the cipherblock is decrypted the result of the decryption is fed into the authentication as if it had been in AuthBody.

That raises the question whether the encrypted payload an AuthBody or the embeeded bytes (Authbody.payload). Makes sense to be specific about that.

proto/auth/auth.proto
```diff
@@ -18,9 +19,14 @@ message AuthContainer {
...
   org.lfedge.eve.common.HashAlgorithm algo = 2;        // hash algorithm used by sender Cert
   bytes   senderCertHash = 3;    // N bytes in length, 1st N bytes of sender siging cert sha256 hash
   bytes    signatureHash = 4;    // signature of the sha256 hash of the payload
   bytes    senderCert = 5;       // full senderCert needed for some payloads
+  org.lfedge.eve.common.CipherContext cipherContext = 6; // context for decrypting payload if defined
+  org.lfedge.eve.common.CipherBlock cipherData = 7; // encrypted payload if defined
```
eriknordmark, line 31: For current encryption we have the CipherContext and Cipherblock in different parts of the API and the CipherBlock references the CipherContext using the cipherContextId. Do we expact to do that here, or will the contextId be empty and the reference is implicit by the contect and block being in the same AuthContainer?

deitch:
I cannot comment on the content of the change (especially as Erik did that). I can comment on the toolset. It looks like yo ran this using a single step more recent minor version of `protoc`, but a several steps older minor version of `protoc-gen-go`.

Can you rerun the `make proto` with the right tools? Or more easily, inside the devcontainer? If you are in vscode, it should recognize it for you. Else follow the instructions in the README.

proto/auth/auth.proto
rouming: ContextID should be valid for CipherBlock and CipherContext, in the decryption routine I check that those two should match. So there is no any difference here.

rouming:
> I cannot comment on the content of the change (especially as Erik did that). I can comment on the toolset. It looks like yo ran this using a single step more recent minor version of `protoc`, but a several steps older minor version of `protoc-gen-go`.
> 
> Can you rerun the `make proto` with the right tools? Or more easily, inside the devcontainer? If you are in vscode, it should recognize it for you. Else follow the instructions in the README.

Avi, as I said in the mail thread I can't install devcontainer. I spent quite a lot of time trying to build it from sources and also failed (I'm on linux, not on mac). I'm not using vscode. The whole procedure of 'make proto' now is inconvenient for me and I expect more complains will follow (the whole eve team is using linux, not mac). 

I can rebuild the protoc-gen-go, but as I said this is very inconvenient.

proto/auth/auth.proto
rouming: Will re-word it. Let's see how it goes.

proto/evecommon/acipherinfo.proto
```diff
@@ -3,11 +3,14 @@
 
 syntax = "proto3";
 
-import "evecommon/evecommon.proto";
+package org.lfedge.eve.common;
+
+option go_package  = "github.com/lf-edge/eve-api/go/evecommon";
+option java_package = "org.lfedge.eve.common";
+option java_multiple_files = true;
+option java_outer_classname = "ACipherInfo";
```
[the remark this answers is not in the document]
rouming: Actually java options are defined for the whole evecommon package. So I will keep it for the acipherinfo as well just to be similar to what already defined in the evecommon folder.

rouming:
@deitch @eriknordmark updated, please take a look.

proto/auth/auth.proto
eriknordmark: OK

eriknordmark (approved):
LGTM but if you can pull in https://github.com/lf-edge/eve-api/pull/15 you should be able to manually use the devcontainer on Linux.

rouming:
> LGTM but if you can pull in #15 you should be able to manually use the devcontainer on Linux.

Thanks, @eriknordmark . These changes work for me. I replied here: https://github.com/lf-edge/eve-api/pull/15/commits/3ca855b339c1a6f5e26deb3730472be45ca9d90c#r1298131945

PR #15, opened 2023-08-17
Fix manual devcontainer workflow on Linux

eriknordmark:
This ensures that devcontainer can be built and manually run.

Fixes issue #11 and issue #16


README.md
```diff
@@ -103,6 +103,21 @@ devcontainer --workspace-folder . up
...
+#### Manual ad-hoc use of devcontainer on Linux
+
+First the devcontainer needs to be built specifying the TARGETOS:
+```bash
+docker build -f .devcontainer/Dockerfile --build-arg TARGETOS=linux .
+```
+
+Then the produced container (hash) can be invoked and make run:
+```bash
+docker run --network=host -it -v $PWD:/home d9ca10a77a8c
```
deitch, line 115: Some suggestions:

1. Can we make this a single command? 
2. We shouldn't need `network=host`, as it by default has bridged internet access.
3. This may make the files owned by root.
4. Give a name to the image?
5. Why does the TARGETOS need to be linux? It doesn't particularly care, but that is what it is by default anyways.
 
So perhaps:

```bash
$ docker build -f .devcontainer/Dockerfile -t eve-api-builder .
$ docker run -v $(pwd):/src -w /src -u $(id -u) eve-api-builder make proto
```

deitch (commented):
Suggestion

README.md
eriknordmark: I'll try those commands.

TARGETOS is blank by default (for some unknown reason) which resulted in a 404 error when download.

deitch: Ah so set a default in the dockerfile.

rouming: That is a good change. I ran Avi's commands, because the last one calls 'make proto'. I suggest the following changes:

1. Make TARGETOS=linux by default:
```
[roman@finn eve-api]$ git diff .devcontainer/
diff --git a/.devcontainer/Dockerfile b/.devcontainer/Dockerfile
index 6bc4a3f4cf06..9ef638328940 100644
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,7 +1,7 @@
 # syntax=docker/dockerfile:1.6.0
 FROM alpine:3.18
 
-ARG TARGETOS
+ARG TARGETOS=linux
 ARG TARGETARCH
 ARG PROTOC_GEN_GO_VERSION=1.31.0
 ARG PROTOC_VERSION=23.4
```

2. Add '--rm' to the 'run' command to delete started container:
```
docker run --rm -v $(pwd):/src -w /src -u $(id -u) eve-api-builder make proto
```

Then everything works! Thanks all.

rouming: And why can't we include these two commands into the makefile under the 'make proto' target? Then no readme and everyone will be familiar with the procedure by calling 'make proto' from the eve-api folder:

```
[roman@finn eve-api]$ git diff Makefile
diff --git a/Makefile b/Makefile
index db775d116082..2efe11298ce6 100644
--- a/Makefile
+++ b/Makefile
@@ -4,9 +4,9 @@ proto-diagram:
        dot ./images/devconfig.dot -Tsvg -o ./images/devconfig.svg
        echo generated ./images/devconfig.*
 
-.PHONY: proto-api-%
+.PHONY: proto-api-% proto gen-proto
 
-proto: go python proto-diagram
+gen-proto: go python proto-diagram
        @echo Done building protobuf, you may want to vendor it into your packages, e.g. pkg/pillar.
        @echo See ./go/README.md for more information.
 
@@ -19,3 +19,6 @@ proto-api-%:
        rm -rf $*/*/; mkdir -p $* # building $@
        protoc -I./proto --$(*)_out=$(PROTOC_OUT_OPTS)./$* proto/*/*.proto
 
+proto:
+       docker build -f .devcontainer/Dockerfile -t eve-api-builder .
+       docker run --rm -v $(PWD):/src -w /src -u $$(id -u) eve-api-builder make gen-proto

```

deitch: > And why can't we include these two commands into the makefile under the 'make proto' target? Then no readme and everyone will be familiar with the procedure by calling 'make proto' from the eve-api folder

Because then it won't work inside devcontainers. The idea is to have a standard defined way of saying, "it should run under this spec container", so that calling `make proto` happens automatically inside the container.

If you want a different target or variable control, e.g. `make proto CONTAINER=true` or `make proto-container` (which means, "run all of the container build and setup stuff and run the command inside")  then we can do that. I think the clean and simple `make proto` should remain the "do it native" call, and another call (or var) can wrap it.

The real goal is to get rid of any of that container commands from the Makefile, but it looks like the devcontainer stuff still has work to do to get there.

rouming: Sorry, Avi, I strongly disagree. Calling make IS a standard defined way.

README.md
```diff
@@ -103,6 +103,14 @@ devcontainer --workspace-folder . up
 devcontainer exec --workspace-folder . make proto
 ```
 
+#### Manual ad-hoc use of devcontainer on Linux
+
+First the devcontainer needs to be built and then run to do the make proto:
+```bash
+docker build -f .devcontainer/Dockerfile -t eve-api-builder .
+docker run -v $(pwd):/src -w /src -u $(id -u) eve-api-builder make proto
```
rouming, line 111: Each run of this command creates a container and does not delete it on command exit. What if to add '--rm' as I noted in the comment?

eriknordmark: Sorry I forgot to push the latest yesterday.

rouming (approved):
Ack.

README.md
deitch: Yes, it is, but it is not standard to call a docker wrapper or not to call a docker wrapper. We want the clean, simple way to be the simplest command, i.e. `make proto`. If you want a container to build it for you, that is an extra layer.

I understand where you are coming from; I used to do all of my Makefiles with calls to docker, and you passed a variable to override it, usually something like: `make all BUILD=local` or similar. Actually, I think an early eve Makefile had just that. But it gets really messy really quickly (our eve Makefile is awfully complex, especially where we try to build the docker builder).

The simplest and most understandable it:

* `make proto` - I will build with whatever tools I find, I assume they are here. Whether you have them locally, or call them via a docker container, or have some tool that installs them, or use devcontainers (which I really hope matures soon, although the VS Code and GitHub workspaces implementations are pretty good), I will just take it. Lowest and simplest.
* `make proto BUILD=docker` (or `make proto-docker`, etc) - I will do the build by wrapping it with some builder layer, which means I will call the builder layer, and then call the simple `make proto` inside there.

eriknordmark: Does this automated build discussion belong in this manual build PR? How do we move forward?

deitch: I think it does. In the end, we are just debating if:

- `make proto` calls the commands and `make proto BUILD=container` (or `make proto-container`) wraps it with `docker run`; OR
- `make proto BUILD=local` (or `make proto-local`) calls the commands and `make proto` calls the container and wraps it with `docker run`

I prefer the first, based on our experiences with eve; Roman prefers the second.

rouming: I prefer just to call `make proto`, but how it works under the hood it does not matter, until I don't need to build/install any new tools (devcontainer, vscode, etc).

deitch: Well, you already have to install docker, so already _some_ tools.

I know, I understand. The principle still applies: the most basic command - the one that will be executed when all prerequisites are in place, i.e. when you are _inside_ your ready-to-go build environment (a container, local with all tools, devcontainer via CLI or vscode, whatever) - is the simplest command, so that needs to be `make proto`.

If you are _outside_ such an environment and need to launch it, and want a make target for that (which I understand and agree with completely), then you need a higher-level command, so one of:

* `make proto-container`
* `make proto BUILD=container`

It doesn't matter which (pick one), as long it is wrapping the basic command, so that the simplest command (`make proto`) is run _inside_ your ready-to-run dev environment.

Just pick one, put it in, let's do this.

deitch (approved):

deitch:
yes, but I do not have merge rights.

eriknordmark:
> yes, but I do not have merge rights.

Odd. But now you do.

PR #22, opened 2023-08-31
Introduce compound edge device config message

rouming:
New `CompoundEdgeDevConfig` message includes regular edge config packed into the `AuthContainer`,   
corresponding node/applications commands and radio config.
    
The message will be used for the LOC case, when there is only one "/config" endpoint and LOC can't modify the original edge config, but various device or application commands should be delivered and processed.
    
This approach is similar to LPS (local profile server) case and the only difference is that one message combines all possible commands in one structure simplifying implementation.

CC: @milan-zededa


eriknordmark (approved):
Please add a section to apiv2.md about the new message and its usage

rouming:
Difference to the previous version:
* Moved CompoundEdgeDevConfig declaration to the config/compound_devconfig.prot 
* Add a few doc tweaks.

@eriknordmark please take a look.

eriknordmark (approved):
Looks good 

In the LOC case do we require https? And does the controller specify the server certificate to trust for LOC now that it includes the LPS commands?

If not i think we need to add that as a follow-on PR

rouming:
> Looks good
> 
> In the LOC case do we require https? And does the controller specify the server certificate to trust for LOC now that it includes the LPS commands?
> 
> If not i think we need to add that as a follow-on PR

Now we use http only, because not clear how to create certificates for customers which can be recognized by eve. So this is unsolved for now.

PR #24, opened 2023-09-04
[DO NOT MERGE] test yetus checks on proto files

eriknordmark:
This should result in errors from the yetus workflow.


eriknordmark:
Odd that it doesn't re-run yetus when I push here. Worked in a branch.

Issue #20, opened 2023-08-24
workflow checking for diffs fails even if no diffs


eriknordmark:
The check for differences seems to result in exit 1 even if there are no changes (see below), and reading the code it will definitely exit 1 if there are tool version changes.

0s
Run echo "changed and uncommitted files, excluding versions"
changed and uncommitted files, excluding versions

differences other than tool version changes:

differences including tool version changes:
Error: Process completed with exit code [1

From https://github.com/lf-edge/eve-api/actions/runs/5960807127/job/16168787320?pr=19#step:4:1

rene:
Issue is already fixed by PR #25 .

PR #25, opened 2023-09-04
Improve workflows

eriknordmark:
Seems like the diff check in ci.yaml is using a constant and not a variable. Fixes issue #20 
And we want to exclude the images from the comparison since the buider produces different output each time it is run with identical source files.

Second commit was just checking yamllint and discovering that the indentation etc is off for yetus.yaml.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,5 +1,7 @@
+# yamllint disable rule:line-length
+---
 name: Build and Test
-on:
+on:  # yamllint disable-line rule:truthy
   pull_request:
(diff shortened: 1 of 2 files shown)


deitch:
> And we want to exclude the images from the comparison

Yes, it is a frustration of mine. I have reached out to the maintainers, but have not heard back. I dug into how the images are generated, and it is fixable, but we are blocked until they respond. I will ping them again.

I kept the image check under the assumption that we would fix it quickly, and thus better to be sure it was changed. Clearly, it is not happening quickly, so your approach is better.

PR #26, opened 2023-09-05
Fix typo in patch envelope

uncleDecart:
This PR fixes typo introduced in #12.

**Note:** using `devcontainer` to run `make proto` generated big diff because `protoc-gen-go` version was bumped. @deitch, please confirm :D 

cc: @eriknordmark


deitch (approved):
LGTM.

Hopefully, as we standardize on the container now, future diffs will be simpler.

PR #27, opened 2023-09-05
Implement acknowledgement for the compound config

rouming:
This PR adds 'timestamp' and 'last_cmd_timestamp' fields.  New fields have the following meaning (copied from the local_profile.proto):
  
```  
     A list of commands requested to be executed for locally running application instances.
     A new request created for the same application should overwrite the previous entry
     with the 'timestamp' field updated. In other words, the list should contain at most
     one entry for each application instance.
     It is not required for the Local profile server to persist command requests.
     Also, it is not required for the Local profile server to stop submitting command
     requests that have been already processed by EVE. Using the `timestamp` field,
     EVE is able to determine if a given command request has been already handled or not.
     To check if the last requested command has completed, compare its timestamp with
     'last_cmd_timestamp' from `LocalAppInfo` message, submitted by EVE in the request
```
For the compound config message I change the order of all fields to have the sequence field to be the first. This is safe, because the structure is not in use anywhere.


proto/config/devconfig.proto
```diff
@@ -170,6 +170,9 @@ message EdgeDevConfig {
 message ConfigRequest {
   string configHash = 1;
   bytes  integrity_token = 2; // value provided by controller during remote attestation
+  // Value of the field `seq` from the last received and successfully processed
+  // `CompoundEdgeDevConfig`.
+  uint64 ack_seq = 3;
```
eriknordmark, line 175: ConfigRequest message goes with the ConfigResponse message.

If you need to POST a message with additional fields and/or semantics then it would make sense to define a new CompoundEdgeDevConfigRequest. For instance, an integrity_token probably doesn't make any sense since there isn't necessarily an end-to-end remote attestation in the LOC case. So the semantics will be more clear with a new request message.

proto/config/compound_devconfig.proto
```diff
@@ -15,8 +15,11 @@ import "profile/local_profile.proto";
 // needed for the LOC case, when a single "/compound-config" endpoint can be
 // used.
 message CompoundEdgeDevConfig {
-   org.lfedge.eve.auth.AuthContainer protected_config = 1;
-   org.lfedge.eve.profile.LocalDevCmd dev_cmd = 2;
-   org.lfedge.eve.profile.LocalAppCmdList app_cmd_list = 3;
-   org.lfedge.eve.profile.RadioConfig radio_config = 4;
+   // Sequence number of compound config, used to guarantee all commands
+   // are processed. See the pair field `ack_seq` of the `ConfigRequest`.
```
eriknordmark, line 19: I don't understand how this will be used to ensure that all commands are processed.
What are the assumptions about when the device will increment the number in the request e.g., will it do that after it has timed out connecting to the LOC? After a reboot?

A sequence number (or a random nonce) can be used to match a request with a response, but if the response is to a POST then the request (POST payload) and response to POST are naturally associated.
So not clear what problem you are trying to solve.

rouming: The logic is similar to the LPS 'timestamp' and 'last_cmd_timestamp' which goes along with each command or info request (see the local_profile.proto). The word timestamp in case of LPS looks clumsy to me, and 'seq/ack_seq' is the default term for the acknowledgement mechanism.

I'm trying to achieve acknowledgement of processed commands, so LOC knows what has been processed, what not. Probably the statement 'guarantee all commands are processed' is incorrect here and needs re-wording. I will change this.

milan-zededa: Note that the reason we called the field `timestamp` was just to make it less bug prone for the developers of LPS (which is not us in that case). For correct acknowledgement of processed commands, it is required that two successive but distinct requests made will have different timestamps (aka seq. numbers) attached (mentioned [here](https://github.com/lf-edge/eve-api/blob/main/proto/profile/local_profile.proto#L113-L121)). This needs to be ensured also across LPS reboots. Simple non-persistent counter does not satisfy this requirement. Storing e.g. unix timestamp into `seq` will work. So we decided to call it `timestamp` to simply give hint to LPS developers how to develop this.
In case of LOC you are in control of its implementation so you can chose better terminology.

rouming:
Difference to the previous version:
* Create a separate message `CompoundEdgeDevConfigRequest` which includes the `ConfigRequest` and the `ack_seq` field.
* Code comments tweaks.

proto/config/compound_devconfig.proto
rouming: Thanks @milan-zededa for the clarification!

rouming: just to be clear: I clearly understand why you chose the timestamp and I did not want to raise any hot discussion with my "clumsy terminology" statement (although I don't take it back) :) I just  don't like the timestamp approach, because this does not guarantee you the persistence - clocks can drift in the past as well (simple scenario: no access to the internet for weeks, all good, command is issued, LPS vm reboots, connection to the internet appeared, LPS syncs time with NTP, time goes to the past, boom). Yes, I agree, much better than counter, but I don't like this "kinda works" things.

BTW, the LOC can also be rebooted. But. The problem can be easily fixed by the LOC implementation (as you said): if `ack_seq` (comes from EVE) is detected > than `seq`, then LOC can update the `seq` and set it to the max. 

Anyway. Thanks for the clarification once again.

eriknordmark: > The logic is similar to the LPS 'timestamp' and 'last_cmd_timestamp' which goes along with each command or info request (see the local_profile.proto). The word timestamp in case of LPS looks clumsy to me, and 'seq/ack_seq' is the default term for the acknowledgement mechanism.

sequences numbers tend to imply monotonically increasing (without gaps) numbers and potentially the need to handle wraparound from 2^N-1 to zero.
timestamp is a somewhat predictable thing.
A nonce would be a random thing which can't be guessed.
Not clear what the semantics are other than the LOC echoing the most recently received number. Can/should the LOC reject "old" sequence numbers? The choice of name would make me think that is part of the semantics. But that implies that EVE has to remember the largest sequence number it has ever used.

eriknordmark: And the suggestion to have the LOC not persist the largest number it has used but instead learn it from EVE's seq_ack sounds like all the semantics you get is that of a non-random nonce; you effectively can only check for equal since greater/less doesn't mean anything if you adjust the sequence number based on the seq_ack.

Thus you have a timestamp from some form of virtual clock and nothing more AFAICT. So might as well use a timestamp from the wall clock.,

If you really need and want to check for greater than, then you need something a lot more robust than what you are describing.

proto/config/compound_devconfig.proto
```diff
@@ -8,15 +8,27 @@ option go_package  = "github.com/lf-edge/eve-api/go/config";
 option java_package = "org.lfedge.eve.config";
 
 import "auth/auth.proto";
+import "config/devconfig.proto";
 import "profile/local_profile.proto";
 
+message CompoundEdgeDevConfigRequest {
+  // Value of the field `seq` from the last received and successfully processed
```
eriknordmark, line 15: Nit (and I know whether we have the same wording in the LPS  API) but "processed" might not be what we do here.
Taking the most extreme example, if the LOC sends the poweroff command, then "processed" would mean that it sends the seq_ack for that command when the device is in fact powered off!

It has received it and will act on it (with sufficient persistence in EVE to act later even in precense of some disruption - which is why we persist e.g., the purge counter for an app instance to remember to do that after a power outage if we didn't get it done before the power went out.)

proto/config/compound_devconfig.proto
```diff
@@ -8,15 +8,27 @@ option go_package  = "github.com/lf-edge/eve-api/go/config";
...
+  // `CompoundEdgeDevConfig`.
+  uint64 ack_seq = 1;
+  ConfigRequest cfgReq = 2;
```
eriknordmark, line 18: A buflint issue on this line.

proto/config/compound_devconfig.proto
rouming: In the current implementation, I don't see where "old" sequence numbers can appear and cause any problems. There are two possible scenarios where sequence numbers are out of sync:

a) EVE has been restarted. This case is dealt with in permanent files (already implemented for LPS) where all pending commands are stored with their respective timestamps/sequence numbers. EVE sends `ack_seq` equal to 0, so LOC does nothing because its `seq` number is higher. The next configuration response (sent from LOC) with `seq` number will be processed by the default `ProcessReceived*Commands` functions (all pending commands have been read from file), `ack_seq` number will be set to `seq` .

b) LOC is restarted. EVE requests a configuration with `ack_seq` greater than `seq` LOC number (which is 0 after restart). Since this is the first configuration request since the LOC restart, two scenarios are possible:
b.1) No commands were issued before the first configuration request. The LOC updates its `seq` number with the number it received from EVE, so that all future commands and responses will be tagged with the same `seq` number.
b.2) There were commands sent right after the LOC restarted and before EVE requested its first configuration. These are new commands, so they must be executed on EVE. As soon as the LOC receives the first configuration request with `ack_seq` > `seq`, the LOC sets its own `seq` = `ack_seq` + 1 and updates all issued commands accordingly. In this case, EVE always gets a `seq` number greater than it is currently processing in order to start processing new task commands.

The only case that is missed here is that when the LOC is restarted, all commands issued on the LOC side are lost. I suppose for now this is a small problem that we can live with.

proto/config/compound_devconfig.proto
rouming: I do not think that "command processing" implies "command completion". Command is processed and then in the future completed. I can add a few clarification words about it. What's important, that "processed" means "commands are persistent now in the file", which means that won't be lost, so the `ack_seq` can be safely set to the `seq`.

rouming:
Difference to the previous version:
* rename `cfgReq` to `cfg_req`
* a few code comments tweaks

proto/config/compound_devconfig.proto
eriknordmark: Agreed.

proto/config/compound_devconfig.proto
eriknordmark: While things will work based on how you describe it, calling this a sequence number is misleading and likely confusing (it looks more like a virtual clock if I don't misremember the Lamport paper from the early days of distributed computing).
And having to special case b.2 seems like unneeded extra complexity.
Perhaps there is also a case when both EVE and LOC starts the first time which you need to handle?
In such a case if commands are sent from the LOC with seqno = 0, EVE processes them, sends back seq_ack=1, LOC receives this and updates it seq and sends the same commands again. There is a risk that commands get processed twice in such a case.

I think that if you want to do sequence number semantics you should use something common such as  the TCP SYN exchange where the two ends reliably agree on a pair of sequence numbers and retain state, plus can detect when the peer has restarted and lost state. (There are other reliable transport protocols such as SCTP or QUIC which take a slightly different approach, but my point is inventing a new protocol for sequence number coordinating is likely to lead to new complexity.)

Alternatively use much simpler semantics of a timestamp or random nonce which is echo'ed.
Is there a case where that would be insufficient? Is there a case where the LOC or EVE must reject something which is old which can not be solved with either a random nonce or by a timestamp?

rouming: The ack echoes back the seq, once processed, as you described. So eve sends ack=0 for the first time in request and if there is something to execute, loc sends seq=1, when eve processes the compound config it sends ack=1 back, which means for the loc "commands below or equal to ack are processed". This is similar to timestamp semantics of the loc, but I use monotonic seq numbers because timestamp can jump into the past, this case I described here: https://github.com/lf-edge/eve-api/pull/27#discussion_r1317123004

eriknordmark: I don't see why we need to define a new protocol here, particularly since it is more complex than it should be as case b.2 indicates (and yes, others might think that is a trivial tweak but to me it indicates underlying complexity. Sorry I've seen this too often in the IETF where folks specify protocols for their whole carreer - it is a lot harder than one thinks.)

I feel if we so this way we need a protocol specification (akin to [RFC 793](https://www.ietf.org/rfc/rfc9293.pdf)) which specifies how the sequence numbers are coordinated between the two end points.

What happens when the sequence numbers are out of whack and the LOC sends a seq which EVE will ignore? Do we need to define an error? Otherwise that will be a silent failure to act on the command.

I also think that in this approach there should be an explicit SYN flag to synchronize when the LOC has forgotten the last sequence number.

Timestamps have none of this complexity; merely assumes that folks run ntp (and it you don't then clocks will drift and downloads from AWS/Azure will fail.)

rouming: What's interesting, the doc to the `timestamp` field says the following:
```
   // Timestamp to record when the request to run the command was made.
   // The format of the timestamp is not defined. It can be a Unix timestamp
   // or a different time representation. It is not even required for the timestamp
   // to match the real time or to be in-sync with the device clock.
   // What is required, however, is that two successive but distinct requests made
   // for the same application will have different timestamps attached.
   // This requirement applies even between restarts of the Local profile server.
   // A request made after a restart should not have the same timestamp attached
   // as the previous request made for the same application before the restart.
```
https://github.com/lf-edge/eve-api/blob/26edd90857ba8c8c2122cdc0a16f05a9ef569a77/proto/profile/local_profile.proto#L113

It says ` The format of the timestamp is not defined. It is not even required for the timestamp to match the real time ... ` , and `What is required, however, is that two successive but distinct requests .. will have different timestamps attached`  - so having a monotonic sequence number I still hold these requirements and guarantee that LOC never sends "old" sequence (which is not guaranteed for the timestamp).

rouming: Anyway, please send me your NAK for the sequence number and confirmation that we use timestamp regardless the described issue and I will quickly update PRs.

rouming:
Difference to the previous version:
* Rename seq/ack_seq fields to timestamp/last_cmd_timestamp as requested by Erik

proto/config/compound_devconfig.proto
eriknordmark: Sorry for being slow in responding; been a very busy week.

The timestamp/nonce approach works as long as the LPS/LOC only has one outstanding command at any given time - basically needs to repeat that until EVE acks it.

Doing anything fancier would imply a subset of a TCP-like protocol with synchronization of sequence numbers, retransmissions (resend the sequence numbers which has not yet been ack'ed), etc.

Down the road we can explore making sure that these commands are all declarative, meaning that the set of commands in a single compound config can be extended if the previous set has not been ack'ed yet. We can talk about this later.

eriknordmark (approved):
LGTM.

AS you said, since there is no LOC implementation using CompoundEdgeDevConfig yet we can ignore the bufcompat errors.

PR #29, opened 2023-10-05
Fix Patch Envelope status spelling error and add ZiOpaqueAppInstanceStatus  

uncleDecart:
We need ZiOpaqueAppInstanceStatus in order to distinguish between status send by app and information about patch envelope assigned to app instance


uncleDecart:
@eriknordmark can you please check?Renaming is a breaking change, but status handling has not yet been merged to eve, so it does not affect it.

eriknordmark (approved):
Since this is not used by any code we know of in EVE or controllers, fixing the typo in PATCH_RECIEVED is not an issue even at recompile of that software. 
And since we are not changing the value it would never be an issue in the protobuf wire encoding compatibility.

Ignoring the yetus complaints about the prefixes since we follow the patten for the existing enum values.

eriknordmark:
Looking at the build and test failure, the diffs are additional empty lines in the output from protoc. Are you building this using the build container (which runs a well-know version of protoc)?
In any case we'll merge it, but want to understand why we get false alarms like this.

uncleDecart:
@eriknordmark I rebuild it using devcontainer, works like charm :)

PR #33, opened 2023-10-30
Patch Envelopes: Flatten ZInfoPatchEnvelopeUsage and remove opaque status

uncleDecart:
Before submitting PR with Patch Envelopes status I'm opening one for EVE API to flatten `ZInfoPatchEnvelopeUsage` and remove Opaque Status since it's duplication of AppInstMetadata 

Since this slightly changes API @gkodali-zededa @udit-zededa please review and comment :)


proto/info/patch_envelope.proto
```diff
@@ -55,28 +45,15 @@ message ZInfoPatchEnvelope {
...
     // Version of PatchEnvelope
     string version = 2;
-    // Number of times this PatchEnvelope
-    // was sucessfully downloaded
-    uint64 timesDownloaded = 3;
+    // count the number of times app instance called patch APIs
+    uint64 patchApiCallCount = 3;
+    // count the number of times app instance actually downloaded
+    // whole patch envelope or part of it
+    uint64 timesDownload = 4;
```
uncleDecart, line 58: Apart from breaking changes, yetus suggests to change this to snake case. I didn't know that we are using snake case in proto messages

eriknordmark: Yes, we are using snake case in new protobuf (unless it is new values in an existing enum when we follow the existing convention for that enum).

The protoc produces the same output i.e., lowerCamelCase even as you change the proto file to use lower_snake_case.

proto/info/info.proto
```diff
@@ -1135,7 +1134,6 @@ message ZInfoMsg {
     ZInfoHardware hwinfo = 18;
     ZInfoLocation locinfo = 19;
     ZInfoPatchEnvelope patchInfo = 20;
-    ZInfoOpaqueAppInstanceStatus opaqueAppInstStatus = 21;
```
eriknordmark, line 1138: I think it would be good to add a comment that 21 was previously defined (even though it hasn't been used) and is now deprecated.

proto/info/patch_envelope.proto
eriknordmark, line 58: Both of them have the semantics of "counter", but the naming convention does not tell that at all - "times" vs. "count". Can we make the names indicate that both are counters?

eriknordmark (changes requested):
Some items to fix, See separate comments.

proto/info/patch_envelope.proto
uncleDecart: Changed to snake_case with count in name

proto/info/info.proto
uncleDecart: Added deprecated comment

gkodali-zededa: https://protobuf.dev/programming-guides/proto2/

According to this `ZInfoOpaqueAppInstanceStatus opaqueAppInstStatus = 21 [deprecated=true];` is the suggested way to deprecate existing fields. Can we start doing that?

proto/info/info.proto
```diff
@@ -71,7 +71,6 @@ enum ZInfoTypes {
   ZiEdgeview = 12;
   ZiLocation = 13;
   ZiPatchEnvelope = 14;
-  ZiOpaqueAppInstanceStatus = 15;
```
eriknordmark, line 74: Sorry I missed this one. Makes sense to put a // deprecated = 14;
comment here as well.

proto/info/info.proto
eriknordmark: @gkodali-zededa Thanks for that pointer. 
But reading that document it seems to say that reserved 21;
is the better approach since deprecated doesn't prevent it from being used (but might result in protoc warnings for some languages.)

@uncleDecart want to try with reserved for this and the field above?
I'm not sure whether there should be a single reserved statement per message with multiple values, or if you can have multiple reserved statements.

uncleDecart: @eriknordmark I added reserved field for things that I've deleted, pb files generated, so I assume it works

eriknordmark (approved):
LGTM

PR #38, opened 2023-12-01
IP routing config for local network instances 

milan-zededa:
These new fields will allow to:
- use DHCP to automatically propagate routes for uplink subnets
  into applications connected to them indirectly through local network
  instances
- configure static IP routes for local network instances and propagate
  them using DHCP to connected applications


proto/config/netinst.proto
```diff
@@ -75,6 +75,13 @@ message NetworkInstanceLispConfig {
   bool experimental = 20;
 }
 
+message IPRoute {
+  // Destination network address in the CIDR format: <IP-address>/<prefix-length>
+  string destination_network = 1;
+  // Gateway IP address
+  string gateway = 2;
```
eriknordmark, line 82: Are we planning to allow some semantics around this being empty/unspecified?
If so we should have a comment about the semantics of an empty string or 0.0.0.0.
If not we should say that this must be set to a valid IP address and can not be all-zeros.

proto/config/netinst.proto
```diff
@@ -107,4 +114,19 @@ message NetworkInstanceConfig {
 
   // static DNS entry, if we are running DNS/DHCP service
   repeated ZnetStaticDNSEntry dns = 41;
+  // Enable to use DHCP to automatically propagate routes for uplink subnets
+  // into applications connected to them indirectly through local network instances.
+  // This option is only valid for local network instances. For other types
+  // of network instances, it will be ignored.
+  bool propagate_connected_routes = 42;
```
eriknordmark, line 121: I always suspect that the question about life and everthing had something to do with IP routing ;-)

proto/config/netinst.proto
```diff
@@ -107,4 +114,19 @@ message NetworkInstanceConfig {
...
+  // List of IP routes statically added to the network instance routing table.
+  // Statically routed subnets are also propagated to connected applications
+  // using DHCP, with gateway set to the network instance bridge IP if it is
+  // outside of the network instance subnet.
+  // IP route gateway may point to an external endpoint (provided that network
+  // instance is not air-gapped), or to an IP address of one of the applications
+  // connected to the network instance.
+  // This option is only valid for local network instances. For other types
+  // of network instances, it will be ignored.
+  repeated IPRoute static_routes = 43;
```
eriknordmark, line 131: Will we allow setting witn the propagate flag and providing static routes?
Makes sense to add to the comment whether this is allowed or not.

naiming-zededa: from some of the previous discussion, we don't allow those automatically, user has to define them

proto/config/netinst.proto
naiming-zededa: Should be do this with 'disable_propagate_connected_routes'? since default case is most people want.

proto/config/netinst.proto
eriknordmark: Sorry, had an imporant typo on my comment.

naiming-zededa: i think they can work independently, good to comment on this api.

naiming-zededa:
we can also introduce a new api-capability for this feature, otherwise user configured static-routes in the cloud side and wondering why it does not show up on the device.

proto/config/netinst.proto
milan-zededa: I'm also considering renaming this to `propagate_uplink_routes`. After all, uplink port may receive additional routes from an external DHCP server. Shouldn't we propagate all the routes associated with an uplink interface?

milan-zededa:
Marking this as draft until some design discussions are closed. Then will also make changes requested in review.

proto/config/netinst.proto
milan-zededa: Done, comment extended.

proto/config/netinst.proto
milan-zededa: Not going to rename, we agreed internally that only connected routes will be propagated, not all routes received from external DHCP server.

proto/config/netinst.proto
milan-zededa: Done, comment extended (empty IP not allowed).

milan-zededa:
> we can also introduce a new api-capability for this feature, otherwise user configured static-routes in the cloud side and wondering why it does not show up on the device.

Done, added new API capability.

milan-zededa:
@eriknordmark, @naiming-zededa PR is ready for re-review.

proto/config/netinst.proto
eriknordmark: Makes sense to keep the name and semantics (need to set to true to change from current behavior).
@naiming-zededa Even if most cases (where app instances have multiple adapters) could benefit from having this set it is a bit scary to start doing that as part of an update by changing the default and have a disable field.

eriknordmark (approved):
LGTM, but as we add this should we also add something about propagate/inject default route or not in the API?
Today we do that by default and can disable it by setting the gateway (for the network if I don't misremember - or is it on the network instance?)

If nothing else, how about documenting that near the comment about static_routes?

milan-zededa:
> LGTM, but as we add this should we also add something about propagate/inject default route or not in the API? Today we do that by default and can disable it by setting the gateway (for the network if I don't misremember - or is it on the network instance?)
> 
> If nothing else, how about documenting that near the comment about static_routes?

Done, added comments explaining default route propagation.

proto/config/netinst.proto
naiming-zededa: Sure.

PR #39, opened 2023-12-13
Move pe app usage to metrics

uncleDecart:
PatchEnvelope AppInst usage is metric, not info. In order for us to
provide consistent updates on patch envelope app usage, we will have to
send information about app which can increase network load.

In contrast, metrics send could be configured on how often they will be sent


proto/info/info.proto
```diff
@@ -930,8 +930,7 @@ message ZInfoApp {
   repeated ZInfoNetwork network = 16;       // up/down; allocated IP
   repeated string volumeRefs = 17;          // volume UUIDs
   repeated ZInfoSnapshot snapshots = 18;    // optional field, used to send list of snapshots on device
-    // optional; information about usage of every patchEnvelope assigned to app
-  repeated ZInfoPatchEnvelopeUsage patchEnvelope = 19; 
+  // deprecated = 19; struct was removed
```
eriknordmark, line 933: How about using
 reserved: 19; // deprecated

uncleDecart: Will update this and yetus complaints

proto/metrics/metrics.proto
```diff
@@ -373,6 +373,21 @@ message appMetric {
   repeated appContainerMetric container = 7;
 
   AppMemoryMetric appMemory = 8;
+
+  repeated appPatchEnvelopeMetric patchEnvelope = 9;
```
eriknordmark, line 377: This is under the appMetric. Do we care about reporting any metrics for patch envelopes which are not (yet) associated with any app instances?

eriknordmark (approved):
LGTM assuming the removed message hasn't yet been used by any controller.
I guess EVE doesn't send the old one so this would just be a compile time issue even if code has been written in some controller?

uncleDecart:
>I guess EVE doesn't send the old one so this would just be a compile time issue even if code has been written in some controller?

It does send this info to controller. I have a patch for EVE [here](https://github.com/lf-edge/eve/pull/3664)

proto/metrics/metrics.proto
uncleDecart: Depends on weather we want sidecar container to be an app entity

proto/info/info.proto
uncleDecart: @eriknordmark do you know if field number 10 in this struct is used? I'd also reserved it

uncleDecart: I reserved field number 10 just in case

PR #40, opened 2023-12-14
Clarification on the default route propagation

milan-zededa:
Further clarification on the default route propagation which hopefully makes it clear.

### DIFF
--- a/proto/config/netinst.proto
+++ b/proto/config/netinst.proto
@@ -113,6 +113,12 @@ message NetworkInstanceConfig {
   AddressType ipType = 39;
 
   // network ip specification
+  // If ip.gateway is set to all-zeroes IP, default route will not be propagated
+  // to applications for interfaces connected to this network instance.
+  // Default route propagation is also automatically suppressed when the network
+  // instance is air-gapped or when the uplink is app-shared without default route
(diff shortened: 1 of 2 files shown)


proto/config/netinst.proto
```diff
@@ -137,11 +143,14 @@ message NetworkInstanceConfig {
...
-  // (not management) and does not have a default route of its own. In the latter case,
-  // it is possible to enforce default route propagation by configuring a static default
-  // route for the network instance.
+  // propagated to connected applications with these exceptions:
+  //  a) default route propagation is explicitly disabled by setting
+  //     NetworkInstanceConfig.ip.gateway to an all-zeroes IP
+  //  b) network instance is air-gapped (without uplink)
+  //  c) the uplink is app-shared (not management) and does not have a default route
+  //     of its own
+  // In the b) and c) cases, it is possible to enforce default route propagation
```
eriknordmark, line 152: What happens if I add a static default route in case a)?  Do you plan to explicitly check and reject/ignore that in the code?

milan-zededa: Yes, the plan was for the zero gateway to take priority.

eriknordmark (approved):
LGTM

PR #41, opened 2023-12-19
Setting all-zeroes gateway IP for NI is not allowed

milan-zededa:
Previous commit "Clarification on default route propagation" is wrong - EVE does not allow to set all-zeroes gateway IP for network instance (parsing would fail among other things). To suppress default route propagation to apps, NI should be without uplink or the uplink should be configured as app-shared with all-zeroes gateway (in NetworkConfig unspecified GW is allowed).

This is defined in a design document for static and connected routes: https://wiki.lfedge.org/display/EVE/Connected+and+Static+IP+Route


eriknordmark (approved):
Thanks for clarifying this.

PR #42, opened 2023-12-21
Parentassigngrp

christoph-zededa:
add `parentassigngrp` field;

Parent Assignment Group is there to reference the parent assignment group in order to make the device
dependent on a different device.
Currently the concrete reason to do this is to make a usb device dependent on the PCI address the USB
controller is using to provent passthrough of the USB controller in one application while trying to passthrough
a USB device on this controller to another application.


proto/config/devmodel.proto
```diff
@@ -153,6 +153,15 @@ message PhysicalIO {
...
   VfList vflist = 9;
+
+  // parentassigngrp
+  //
+  // Parent Assignment Group is there to reference the parent assignment group in order to make the device
+  // dependent on a different device.
+  // Currently the concrete reason to do this is to make a usb device dependent on the PCI address the USB
+  // controller is using to provent passthrough of the USB controller in one application while trying to passthrough
+  // a USB device on this controller to another application.
+  string  parentassigngrp         = 10;
```
milan-zededa, line 164: What about instead having `parent`, referencing logical label of the parent device?
Seems more generic. In some case (like yours) it can prevent separate assignment, in others it could be useful for other purposes.
Plus assigngrp can be empty (meaning not assignable), which would make such device non-referenceable from child devices.

christoph-zededa: What do you mean with <it can prevent separate assignment>?

`parentassigngrp` allows having several parent devices that are all in the same `assigngrp`; I think that's not possible if we reference the `logicallabel`.

milan-zededa: It is supposed to be a question so it ends with question mark :)

If you have a pointer to the parent device, you can obtain its assigngrp (aka parentassigngrp), which can be used by other devices.
I just think that having a tree-structure of physical devices could be useful for other cases as well, not just assignment restrictions.

christoph-zededa: > What do you mean with ?

github removed part of the text; I wanted to ask: "What do you mean with it can prevent separate assignment?"

milan-zededa: By "prevent separate assignment" I meant your use-case, i.e. preventing to assign USB device and its controller separately. What I tried to say in that sentence is that with `parent` as opposed to `parentassigngrp` we could perhaps cover some more future use-cases, not only related to device passthrough.

christoph-zededa: For different use-cases different parents might be needed.

F.e. in a use-case of load-balancing/fault-tolerance a device might have different parent devices than in the use-case of passthrough.

milan-zededa: True, fair point.
But I think we should further clarify and define the semantics of the parentassigngrp in the comment.
My understanding is that it limits device assignment and allows only these 2 cases: 
- children and all parent devices are assigned into the same app
- children is assigned while all parent devices stay in the host
Correct?

proto/config/devmodel.proto
```diff
@@ -153,6 +153,15 @@ message PhysicalIO {
   // list of Virtual Functions (VF) for given Physical Function (PF)
   // only applies for PHY_IO_TYPE_PHY_IO_NET_ETH_PF
   VfList vflist = 9;
+
+  // parentassigngrp
+  //
+  // Parent Assignment Group is there to reference the parent assignment group in order to make the device
+  // dependent on a different device.
+  // Currently the concrete reason to do this is to make a usb device dependent on the PCI address the USB
+  // controller is using to provent passthrough of the USB controller in one application while trying to passthrough
```
eriknordmark, line 162: typo "provent"

proto/config/devmodel.proto
eriknordmark: @milan-zededa do we currently require logicallabel to be set? I recall some code which might fallback to using ifname if the logicallabel is not set.

I can see some utility in being able to describe the bus/assignment hierachy for other use cases. The main thing we need to be able to describe for the assignment restrictions is that if a device has a parent, then the device and its parent can not be assigned independently. BUT some parents (the PCI root bridge being the basic example) do allow independent assignment subject to the iommu group constraints. And I don't know if we can have cases of a bus hierarchy which is PCIcontroller-USBbus-PCIcontroller-PCI device, and the grandchild PCI controller might support separaate iommu groups.

christoph-zededa: > we can just as well let EVE to determine if the given assignment is valid

I fear it is not possible in case we use ACS override to do passthrough of PCI devices to different apps even if they're in the same iommu group, but I might be wrong as my knowledge about ACS override is just too small.

eriknordmark: I think it makes more sense to have the controller dictate the rules for assignment than have EVE do something more dynamic, since a dynamic behavior (based on the iommu groups) might change over time with different releases.

milan-zededa: Alright, let's use `parentassigngrp` as proposed by the PR then.
However, I still think we should describe the semantics of the field in the comment rather than just give one concrete example.
See one of my previous comments:
> But I think we should further clarify and define the semantics of the parentassigngrp in the comment.
My understanding is that it limits device assignment and allows only these 2 cases:
>  -  children and all parent devices are assigned into the same app
>  -  children is assigned while all parent devices stay in the host
>
>    Correct?

eriknordmark: @milan-zededa In addition to what you said above, if the parent stays in the host, then different children can be assigned to different app instances. (And in case we ever have grandparents/grandchildren then the top of the tree needs to be in the host, and below that cut different subtrees can be assigned to different app instances.)

@christoph-zededa can you please add Milan's and the above text to the comment?

proto/config/devmodel.proto
```diff
@@ -153,6 +153,19 @@ message PhysicalIO {
...
+  // parentassigngrp
+  //
+  // Parent Assignment Group is there to reference the parent assignment group in order to make the device
+  // dependent on a different device.
+  // Currently the concrete reason to do this is to make a usb device dependent on the PCI address the USB
+  // controller is using to prevent passthrough of the USB controller in one application while trying to passthrough
+  // a USB device on this controller to another application.
+  //
+  // We only allow these two cases:
+  // - children and all parent devices are assigned into the same app
```
eriknordmark, line 166: "children and their parent" (not all parents) is more accurate.

proto/config/devmodel.proto
```diff
@@ -153,6 +153,19 @@ message PhysicalIO {
...
+  // - children is assigned while all parent devices stay in the host
```
eriknordmark, line 167: typo is/are, but might might want to be a bit more explicit with
 - parent device stays in host and one or more children (in separate assignment groups) are assigned to one or more application instances

christoph-zededa: I changed it to:
```
  // - parent device stays in host and one or more children (in separate
  //   assignment groups) are assigned to one or more application instances
```

proto/config/devmodel.proto
christoph-zededa: I changed it to:
```
// We only allow these two cases:
  // - children and their parent devices are assigned to the same application
  //   instance
```

eriknordmark (approved):
LGTM

PR #46, opened 2024-03-12
appsconfig.proto: Add edge node cluster defines

famleebob:
First pass adding edge node cluster to the API

Based on the design and API in https://wiki.lfedge.org/display/EVE/Edge-Node+Clustering+Design


proto/config/appconfig.proto
```diff
@@ -206,6 +206,15 @@ message AppInstanceConfig {
 
   // PatchEnvelopeRef is used to track all patch envelopes used by the app instance
   PatchEnvelopeRef patchRef = 23;
+
+  // This edge-node UUID for the Designate Node for the Application
+  string DesignatedNodeID = 24;
```
eriknordmark, line 211: For new fields it makes sense to follow the lower_snake_case. Note that the protoc compiler create golang struct members in UpperCamelCase thus the code using those definitions do not need to be changed.

proto/config/edge_node_cluster.proto
```diff
@@ -0,0 +1,37 @@
...
+option java_package = "org.lfedge.eve.config";
+
+message EdgeNodeCluster {
+  // cluster name, in case it has multiple cluster on the same site
+  string ClusterName = 1;
+
+  // cluster UUID
+  string ClusterID = 2;
+
+  // Cluster-Interface
```
eriknordmark, line 17: Please provide an example in the comment e.g., "eth1".

proto/config/edge_node_cluster.proto
```diff
@@ -0,0 +1,37 @@
...
+  string ClusterInterface = 3;
+
+  // the 'cluster-prefix' IP address of the 'Cluster-Interface', 10.244.244.2/28
+  string ClusterIPPrefix = 4;
+
+  // This device is 'seed-server'
+  bool IsSeedServer = 5;
```
eriknordmark, line 24: I thought we wouldn't need this. Or will the controller pick a seed server without the user being involved?

proto/config/edge_node_cluster.proto
```diff
@@ -0,0 +1,37 @@
...
+
+  // This device is an 'Agent' node
+  bool IsAgent = 6;
+
+  // SSIP = Seed-ServerIP address to join during bootstrap. E.g. 10.244.244.1
+  string SeedServerIP = 7;
+
+  // encrypted token string, use edge-node TPM to decrypt
+  string EncryptedClusterToken = 8;
```
eriknordmark, line 33: This needs to be of type org.lfedge.eve.common.CipherBlock to use the ECDH object encryption.

proto/config/edge_node_cluster.proto
```diff
@@ -0,0 +1,37 @@
...
+
+  // Resource label list
+  map <string, string> ResourceLabel = 9;
```
eriknordmark, line 36: Are we going to use this in the first phase? If not is is better to not define it now since we might discover when implementing it in EVE and the controllers that we need something slightly different.
If we've already added it in the API we need to deprecate the old fields and add new ones to ensure protobuf compatibility.

eriknordmark:
> Part 1 - first pass adding edge node cluster to the API
> 
> see [Edge-Node Clustering Design Doc](https://docs.google.com/document/d/1CwPmFwrETkGU2WqUkRJlZ8E8CMiLf0V7Y_7FR34WXf4) March 8, 2024 at 9:54 AM.

It would make sense to have a separate document focusing on the cluster configuration and the API and publish that on the LF-Edge EVE wiki.
Then doing a careful design review of the API additions to make sure we are not defining items before we are going to use them, and also that the descriptions/comments are clear enough for an implementor of e.g., adam/eden to use the new API.

proto/config/edge_node_cluster.proto
famleebob: Done,  changed the comment to
```go
// Cluster-Interface, for example "eht1"
```

proto/config/edge_node_cluster.proto
famleebob: updated

proto/config/appconfig.proto
famleebob: Done

famleebob:
I'll hold off creating the generated files until we have agreement on the modified files.

proto/config/edge_node_cluster.proto
naiming-zededa: Yes, we decided to let controller to automatically pick, w/o user to choose. But the entry is needed.

proto/config/storage.proto
```diff
@@ -208,6 +208,12 @@ message Volume {
   string displayName = 7;  // Optional friendly name echo'ed in info message
   bool clear_text = 8;     // Flag to indicate the volume encryption needed or not
   Target target = 9;       // How the volume will be presented to the instance. At the moment only the AppCustom is used
+
+  // To inform the edge-node if the device receiving this Volume is
+  // responsible for volume creation, convert PVC, or not
+  string DesignatedNodeID = 10;
+  // replicated to devices with UUID array, number range from 0 to 2
```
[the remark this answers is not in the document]
naiming-zededa: this 'repeated string' below is the UUID array.

proto/config/edge_node_cluster.proto
naiming-zededa: Sure. we can do that.

naiming-zededa:
> > Part 1 - first pass adding edge node cluster to the API
> > see [Edge-Node Clustering Design Doc](https://docs.google.com/document/d/1CwPmFwrETkGU2WqUkRJlZ8E8CMiLf0V7Y_7FR34WXf4) March 8, 2024 at 9:54 AM.
> 
> It would make sense to have a separate document focusing on the cluster configuration and the API and publish that on the LF-Edge EVE wiki. Then doing a careful design review of the API additions to make sure we are not defining items before we are going to use them, and also that the descriptions/comments are clear enough for an implementor of e.g., adam/eden to use the new API.

this one is posted last week on lfedge wike: https://wiki.lfedge.org/display/EVE/Edge-Node+Clustering+Design

zedi-pramodh:
@famleebob I think we discussed not to put internal docs on public domain

deitch:
> @famleebob I think we discussed not to put internal docs on public domain

@zedi-pramodh I updated the comments to remove links to internal docs.

Erik is right, though, that we need a public design doc on LF-Edge Wiki or in this repo for a design, so good link to the wiki one.

proto/config/storage.proto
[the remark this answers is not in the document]
deitch: I am confused. Is it a UUID? or a number range from 0 to 2 per the comment?

proto/config/edge_node_cluster.proto
```diff
@@ -0,0 +1,37 @@
...
+syntax = "proto3";
+
+import "evecommon/evecommon.proto";
+import "evecommon/acipherinfo.proto";
+
+package org.lfedge.eve.config;
+option go_package = "github.com/lf-edge/eve-api/go/config";
+option java_package = "org.lfedge.eve.config";
+
+message EdgeNodeCluster {
```
milan-zededa, line 13: Will this be part of EdgeDevConfig?

proto/config/edge_node_cluster.proto
deitch: _This comment disappeared since the field was renamed `is_seed_server` in snake case (which is correct)._

The field `is_seed_server` is a one-time thing, yet it is a permanent part of the message `EdgeNodeCluster`. That makes it a short-lived-usefulness-and-meaning yet perpetually-lingering field. it could get confusing over time.
It also doesn't align as much with the language used for k3s.

Would it help to change the field to `cluster_init`? That aligns with how k3s uses it; "node, initialize a cluster instead of joining a cluster". And perhaps a separate `role` field, with `control-plane` and `worker`? We don't use `worker` yet, but we will.

k3s also ignores `--cluster-init` if it already has a local cluster.

That doesn't solve the lingering field, but it helps a bit.

Similar (but slightly different) question about [this field](https://github.com/lf-edge/eve-api/pull/46/files#diff-75211aa7f801986e27d4a3bdea12bd1c296af83e79750487bf1c7cb20a8c7e1aR32-R33):

```proto
  // SSIP = Seed-ServerIP address to join during bootstrap. E.g. 10.244.244.1
  string seed_server_ip = 7;
```

Is this just during _cluster_ bootstrap? In the future, we will have new nodes join a cluster, they will need to be told where to connect at _node_ bootstrap time. The k3s flag is called `--server`, kubeadm uses the field `apiServerEndpoint`, kubelet uses kubectl's `server` field.

Maybe it should be called `joinServer` or similar?

1. I start a single-node cluster (seed server)
2. I have 2 other nodes join - they don't know anything about seed server, just an IP to join
3. a month goes by
4. I ask 2 more nodes to join the cluster - the IP might be the original seed server, it might be something else entirely; seed has no meaning at this point anyways.

famleebob:
> @famleebob I think we discussed not to put internal docs on public domain

Yes we did, my bad.  Just wanted to tie the work done to a specific description of the API

proto/config/storage.proto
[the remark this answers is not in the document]
naiming-zededa: we meant the size of the array is a number, I think it should be 3, to list all the UUIDs needed for this VI for replication. Unless this VI does not need to be replicated by longhorn, then it can be zero size of array.

proto/config/edge_node_cluster.proto
naiming-zededa: yes

proto/config/edge_node_cluster.proto
naiming-zededa: Ok, I think for the EVE API, this isSeedServer is sort of redundant, in the way that
we have the 'cluster_ip_prefix', and the seed_server_ip items. If the seed_server_ip
matches the IP address of the 'cluster_ip_prefix', then the device can perform the
initial bootstrap functions. Otherwise the device will just do the join the the seed_server_ip address.
In terms of the name, 'join_server_ip' or 'bootstrap_server_ip', or we can think of a better one.

Yes, this node might be down and unavailable later for 4th or 5th node to join, but someone needs to
let the new node to know which IP address to use to join. So, in that sense, this 'join_server_ip' can
possibly change later when there is a new node to join.

The instruction has to come from the controller since the new node can not use the kubernetes resource
before it joins kubernetes.
We/controller either needs to do this 'automatically' or user has to do it manually, but we just decided we don't
want user to do manually at least for the initial cluster bring up.

proto/config/edge_node_cluster.proto
```diff
@@ -0,0 +1,37 @@
...
+  // cluster UUID
+  string cluster_id = 2;
+
+  // Cluster-Interface, for example "eth1"
+  string cluster_interface = 3;
+
+  // the 'cluster-prefix' IP address of the 'Cluster-Interface', 10.244.244.2/28
+  string cluster_ip_prefix = 4;
+
+  // This device is 'seed-server'
```
naiming-zededa, line 26: we can remove this bool is_see_server here.

proto/config/edge_node_cluster.proto
```diff
@@ -0,0 +1,37 @@
...
+  bool is_seed_server = 5;
+
+  // This device is an 'Agent' node
+  bool is_agent = 6;
+
+  // SSIP = Seed-ServerIP address to join during bootstrap. E.g. 10.244.244.1
+  string seed_server_ip = 7;
```
naiming-zededa, line 33: maybe change this item name to server_joinning_ip,
and add comment saying 'During bootstrap, the device has the cluster_ip_prefix IP address the same as the 'server_joinning_ip', it is responsible for the cluster-init operations.' or something like that.

proto/config/appconfig.proto
```diff
@@ -206,6 +206,15 @@ message AppInstanceConfig {
...
   // PatchEnvelopeRef is used to track all patch envelopes used by the app instance
   PatchEnvelopeRef patchRef = 23;
+
+  // This edge-node UUID for the Designate Node for the Application
+  string designated_node_id = 24;
+
+  // If the App allows automatic migration
+  bool auto_migration = 25;
+
+  // a set of ResourceLabels for node selection
```
naiming-zededa, line 216: remove the node_select_labels item for now, the first phase.

proto/config/edge_node_cluster.proto
deitch: That is an interesting approach. Would that cover every use case? If so, then it is good. I personally like it when we are explicit, like, `cluster_init = true`, but that really is just personal preference, so it doesn't matter.

proto/config/edge_node_cluster.proto
eriknordmark: I think Milan was pointing out that in this PR it is not part of EdgeDevConfig hence it is unused.

proto/config/storage.proto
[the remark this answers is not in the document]
eriknordmark: The API shouldn't have a max number. And the comment about number range is confusing.

But are we planning to implement controlling the replicas in the next three months or so?
If not, why do we need any of these storage-specific items.

We can just have the controller specify the members of the cluster (message EdgeNodeCluster).
We can add more detailed control as those needs arise and we know better what should be done in the controller vs. by the EVE cluster autonomously.

Can't the cluster elect who should create the volume/PVC?

eriknordmark:
> > @famleebob I think we discussed not to put internal docs on public domain
> 
> @zedi-pramodh I updated the comments to remove links to internal docs.
> 
> Erik is right, though, that we need a public design doc on LF-Edge Wiki or in this repo for a design, so good link to the wiki one.

I took the liberty to add that link to the PR description.

proto/config/appconfig.proto
naiming-zededa: ok, will remove this.

proto/config/edge_node_cluster.proto
naiming-zededa: Ah, got it. I have added:

message EdgeDevConfig {

...

    // cluster configure

    EdgeNodeCluster cluster;

}
in lfedge wike doc.

We need to add here in this PR.

proto/config/storage.proto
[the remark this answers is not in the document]
naiming-zededa: Likely not in first three months. So it's ok to remove this array. We assume everything is replicated to the only 3 nodes.
Have cluster elect works for now, but i doubt will work properly later. since the VIs can be downloaded first, and Apps can be downloaded later, if we have more than 3 nodes invovled, they better be from the consistent configures from the controller. The schemes should work for now and three months or more later.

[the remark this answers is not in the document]
famleebob: comment and the associated field replica_node_id have been removed

proto/config/edge_node_cluster.proto
famleebob: Added the field to EdgeDevConfig

proto/config/edge_node_cluster.proto
famleebob: changed field name to join_server_ip

proto/config/appconfig.proto
famleebob: removed

proto/config/appconfig.proto
```diff
@@ -206,6 +206,9 @@ message AppInstanceConfig {
 
   // PatchEnvelopeRef is used to track all patch envelopes used by the app instance
   PatchEnvelopeRef patchRef = 23;
+
+  // This edge-node UUID for the Designate Node for the Application
+  string designated_node_id = 24;
```
eriknordmark, line 211: Still not clear we need this in the API as opposed to the cluster selecting one control-plane node to do the download and volume creation. If we do it inside the cluster it means that the cluster can re-select the node should the originally selected one fail. Doing it in the API means that the controller (or the user??) now has to react to the failed node and based on that change the configuration it sends to the nodes.

proto/config/storage.proto
```diff
@@ -208,6 +208,10 @@ message Volume {
   string displayName = 7;  // Optional friendly name echo'ed in info message
   bool clear_text = 8;     // Flag to indicate the volume encryption needed or not
   Target target = 9;       // How the volume will be presented to the instance. At the moment only the AppCustom is used
+
+  // To inform the edge-node if the device receiving this Volume is
+  // responsible for volume creation, convert PVC, or not
+  string designated_node_id = 10;
```
eriknordmark, line 214: See comment about this field for the app instance.

proto/info/info.proto
```diff
@@ -912,6 +912,17 @@ message ZInfoSnapshot {
   ErrorInfo snap_err = 6;
 }
 
+enum ZInfoClusterNodeStatus {
+  Node_Status_INVALID = 0;
+  Node_Status_READY = 1;    // cluster reports our node is ready
+  Node_Status_NOTREADY = 2; // cluster reports our node is ready
```
eriknordmark, line 918: Comments are identical for the two fields.

But how does the state of the node relate to the state of the cluster?
Take the case of a cluster where its config lists node1, node2, and node3.
When e.g., node3 reports READY does it mean that not only it has joined the cluster but also that it sees node2 and node1 as part of the cluster?

What status do we expect to see reported by the nodes if/when one of the nodes fail to join the cluster? Will the other two report it as READY?

These semantics should presumably be described in comments in the proto file.

proto/info/info.proto
```diff
@@ -931,6 +942,9 @@ message ZInfoApp {
   repeated ZInfoNetwork network = 16;       // up/down; allocated IP
   repeated string volumeRefs = 17;          // volume UUIDs
   repeated ZInfoSnapshot snapshots = 18;    // optional field, used to send list of snapshots on device
+
+  // The App in cluster mode is currently running on this edge-node
+  bool cluster_app_running = 20;
```
eriknordmark, line 947: I think I already commented on this field but I don't see the comment any more.
Does "running" imply that 1) this node was selected as the place to run the app instance and 2) the app instance has reached "Online" state? 
What about the cases when the node has been selected but the state of the app instance is e.g., booting or halting, halted, error?

It would be helpful to know the intended use (who is going to do something based on this field? Merely for display? Or will a user/controller somehow react to this value to change things?)
Once we know that the semantics can be documented in the proto file.

proto/info/info.proto
```diff
@@ -1143,6 +1157,8 @@ message ZInfoMsg {
     ZInfoHardware hwinfo = 18;
     ZInfoLocation locinfo = 19;
     ZInfoPatchEnvelope patchInfo = 20;
+    // 21 reserved
```
eriknordmark, line 1160: What was 21 used for?

proto/config/edge_node_cluster.proto
```diff
@@ -0,0 +1,33 @@
...
+
+message EdgeNodeCluster {
+  // cluster name, in case it has multiple cluster on the same site
+  string cluster_name = 1;
+
+  // cluster UUID
+  string cluster_id = 2;
+
+  // Cluster-Interface, for example "eth1"
+  string cluster_interface = 3;
```
milan-zededa, line 20: I wonder if this should reference the [logicallabel of the PhysicalIO](https://github.com/lf-edge/eve-api/blob/main/proto/config/devmodel.proto#L118-L122) (like [SystemAdapter does](https://github.com/lf-edge/eve-api/blob/main/proto/config/devmodel.proto#L67-L70)), instead of referencing the network interface name (as used by Linux) directly.

proto/config/edge_node_cluster.proto
```diff
@@ -0,0 +1,33 @@
...
+
+  // the 'cluster-prefix' IP address of the 'Cluster-Interface', 10.244.244.2/28
+  string cluster_ip_prefix = 4;
```
milan-zededa, line 23: Would this be used for the cluster interface instead of using [Network with IP config](https://github.com/lf-edge/eve-api/blob/main/proto/config/netconfig.proto#L19-L20)?

proto/config/edge_node_cluster.proto
famleebob: is_seed_server has been deleted, and seed_server_ip has been changed to join_server_ip.

proto/info/info.proto
famleebob: Don't know! the `reserved` statement is much earlier in the message definition.  The comment is on the edge of visual noise, added it to highlight a statement that is easy to miss ;)

blame:
```
27069fa6 proto/info/info.proto     (Pavel Abramov           2023-10-30 17:09:38 +0100 1142)   reserved 21;
```
commit record:
```
commit 27069fa6864fd403e6bf2b48c38dd34b17e48768
Author: Pavel Abramov <uncle.decart@gmail.com>
Date:   Mon Oct 30 17:09:38 2023 +0100

    PatchEnvelopes: Flatten `ZInfoPatchEnvelopeUsage` and remove Opaque Status
    
    There is no need for additional structure since there is
    1 to 1 releation. This commit also removes opaque status since it is
    duplication of AppInstMetaData
    
    Signed-off-by: Pavel Abramov <uncle.decart@gmail.com>
```

milan-zededa: Yes, Pavel removed `ZInfoOpaqueAppInstanceStatus opaqueAppInstStatus = 21` in the commit referenced by Bob.

proto/config/appconfig.proto
naiming-zededa: From I understand this longhorn persistent provisioning, this will be used to initial setup who should perform this task. Even later this node fails, this does not change.
Another way to do this, to allow cluster to decide, but in general we need to send down a set of nodes for the replication (even though initially we only have 3 nodes), each VI would need to have some election going on among the nodes. That can also be complex.

naiming-zededa: sorry, i thought this part is the volume instance.

for app instance here. We later may want to allow 'manual' provisioning from the user, so this will be the same scheme, let api to indicate which node should run the app.

proto/config/storage.proto
naiming-zededa: i put the reply in above app-instance comment.

proto/info/info.proto
naiming-zededa: I think this is only for the node to report the status as part of the cluster. We'll need another info or metric message to report the cluster status (more like kubernetes view) w/ an elected reporter on the cluster. But that is more complex and is not part of this PR.
So, this is only saying as a device, my node in this cluster status. If all nodes report 'unknown' then the cluster is down.

proto/info/info.proto
naiming-zededa: This info status, main usage is for App 'migration'. the node is not a designated node for the app (related to above app-instance dnid of the app), but kubernetes cluster migrated this app on to this node. The app is running now on this node. We'll let the API changes from the cloud to indicate now the designated node is on this node. All the app domain operations will work normally on this node. e.g. some 'activated' bool will be set to true on the domainstatus, etc.

proto/config/edge_node_cluster.proto
naiming-zededa: Sure, we should change the comments to use logical-label of the interface.

proto/config/edge_node_cluster.proto
naiming-zededa: That would be independent of the original network config on the cluster interface, if it is already there. It adds another prefix on the interface.

proto/config/edge_node_cluster.proto
deitch: That makes sense, @famleebob . And based on @naiming-zededa's description, this gets sent (or looked at) when a node first joins (or initializes) a cluster? So the possible scenarios that an edge device can have are:

* I am joining an existing cluster as a control plane node - I see `join_ip` and `is_agent=false`, and use that IP to join the cluster using `cluster_token` 
* I am joining an existing cluster as a worker node - I see `join_ip` and `is_agent=true`, and use that IP to join the cluster using `cluster_token` 
* I am initializing a new cluster - I see a `join_ip` and `is_agent=false` and `cluster_ip_prefix=join_ip` so I know to initialize
* I already am part of a cluster - ignore

That makes sense.

proto/config/edge_node_cluster.proto
```diff
@@ -0,0 +1,33 @@
...
+
+  // This device is an 'Agent' node
+  bool is_agent = 5;
```
deitch, line 26: Should this be `is_agent` or `is_worker`? I think maybe `worker`? The `agent`/`server` language is a k3s one; `control plane`/`worker` is generic Kubernetes. We _are_ using k3s here, but that is an implementation detail. I think the API should be generic to Kubernetes.

naiming-zededa: Sure. we can change it to is_worker_node. although we have things like join_server_ip, we don't want to be too kubernetes as join_control_plan_node_ip, and 'master' should not be used these days :-)

deitch: Yeah. I think server is fine though. Lots of people use "server" as thing to talk to, and even refer to "control plane server" or "master server". Besides, the only thing I can join is a control plane server anyways.

proto/info/info.proto
eriknordmark: Can you 1) correct the comment and 2) add the above information as comments.

Also, does it make sense to have a separate value for DOWN vs. using UNKNOWN? Either one works for me as long as this is clearly documented.

proto/info/info.proto
eriknordmark: I think this design with designated is problematic since you depend on the controller to adjust the config based on where the app (or volume) is currently running. What if the controller does not adjust it? Should EVE interpret that as it should be migrated back? (Well, it can't since that would be a race in the eventual consistency model.)

Thus I think the placement needs to be based on policy statements (only run on node X, avoid running on node Y, run on same node as app 2) and not involve the controller reacting to where the app insance is currently running; that is not a declarative API statement.

proto/info/info.proto
eriknordmark: @famleebob I assume that if you accidentally tried to reuse 21, given the reserved statement the protoc compiler should have given an error. Did it not?

eriknordmark (changes requested):
You've done some strange pull+merge from main so now other PRs appear in the diffs.

Note that you should NEVER do merge commits. Instead fetch from main and rebase on that.
And for this particular repo with the protobuf files it makes sense to always keep two separate commits: one with the changes to the proto files and a separate one with the generated files. That makes it trivial to rebase since you can drop the second commit (in git rebase -i for instance), and then invoke make to recreate the generated files once you've rebased the commit with the proto changes on top of main.

So please rebase and create the two commits which are needed.

proto/info/info.proto
naiming-zededa: This is just an 'info' message to report this app now start running on this 'node', it does not involve with any placement decision, since it's already being rescheduled by the cluster itself.
The node does not depend on the controller to change the DNID for the app in configure download, the plan is the have the controller eventually change the DNIS to this node, and the App status on this new node will be sorted of becomes 'legal', and start to report App metrics those sort of stuff. Since the reporting of the metrics depends on domainstatus of the app is fully activated.
I think regardless of how the detail of the process on this node, to report this 'app is running on this node' as an 'info' message is important for controller to know, or for user to know.

proto/info/info.proto
famleebob: Yes, it did.  The `reserved` statement gets lost in the text (at least for me).  I added the comment in its where someone might try to add `21`.  I can remove it if you wish.

famleebob:
I tried to pull in the conflict using a `cherry-pick`; assuming that `git` would track it/identify it/ and ignore it on a rebase.  Didn't work.

Completed the rebase, now at two commits, and a forced push.

proto/config/edge_node_cluster.proto
deitch: This still needs to be addressed

famleebob: Thanks! I missed it yesterday.  Done, now is `is_worker_node`.

deitch (approved):

deitch:
Looks good to me, other than Erik's requests (the multiple commits).

famleebob:
Fixed Yetus errors, commit changes, rebase, build, commit generated files.

Should answer all outstanding requests

deitch:
LGTM, but @eriknordmark  needs to approve.

famleebob:
Quick update to try to fix the Yetus errors, did I shoot too soon?

famleebob:
> You've done some strange pull+merge from main so now other PRs appear in the diffs.
> 
> Note that you should NEVER do merge commits. Instead fetch from main and rebase on that. And for this particular repo with the protobuf files it makes sense to always keep two separate commits: one with the changes to the proto files and a separate one with the generated files. That makes it trivial to rebase since you can drop the second commit (in git rebase -i for instance), and then invoke make to recreate the generated files once you've rebased the commit with the proto changes on top of main.
> 
> So please rebase and create the two commits which are needed.

I committed these changes shortly after requested, can we at least mark this done, or not blocking the merge, please?

proto/config/appconfig.proto
```diff
@@ -210,6 +210,9 @@ message AppInstanceConfig {
   // allow AppInstance to discover other AppInstances
   // attached to its network instances
   bool allow_to_discover = 24;
+
+  // This edge-node UUID for the Designate Node for the Application
+  string designated_node_id = 25;
```
eriknordmark, line 215: FWIW I still think this will cause technical debt since the controller has to be involved in re-selecting a designated node if the previously selected one fails. So approving this holding my nose.

proto/config/edge_node_cluster.proto
eriknordmark: I don't see any update in the name or comment to reflect this.

proto/config/edge_node_cluster.proto
eriknordmark: Agreed.

proto/config/storage.proto
eriknordmark: FWIW I still think this will cause technical debt since the controller has to be involved in re-selecting a designated node if the previously selected one fails. So approving this holding my nose.

proto/info/info.proto
eriknordmark: @naiming-zededa We still need to make it clear what "running" means here, in cases where the node is trying to run it (so the state could be booting) or tried to run it (and it failed so the state is currently haling, halted, or the app instance has an error).

Thus is this "running" as in "trying to run and state is currently online" or "selected" as in trying/intending to run on this node? Please clarify in comments and in the name of the field.

eriknordmark (changes requested):
There are still two outstanding clarification requests, then I will approve this holding my nose.

proto/config/edge_node_cluster.proto
naiming-zededa: we'll add at the end of comment: ", this is the logicallabel of the physicalIO"
is this right?

eriknordmark: Yes.

eriknordmark (approved):
LGTM.
You probably have to redo the generated files commit since I just merged another API adddition.

PR #47, opened 2024-03-13
Add allow to discover

uncleDecart:
This flag will allow to expose app instance IP address
via metadata server. Could be useful when you have certain
topology and don't want to have static IP routes, i.e.
sidecar applications.


eriknordmark (commented):
I thought the idea was to only have setting on the app instance to allow discover. Am I confused?

uncleDecart:
> I thought the idea was to only have setting on the app instance to allow discover. Am I confused?

We can move it to app instance level, but this way we can give access granularly, i.e. expose specific port of app instance

eriknordmark:
> > I thought the idea was to only have setting on the app instance to allow discover. Am I confused?
> 
> We can move it to app instance level, but this way we can give access granularly, i.e. expose specific port of app instance

The initial proposal was to have it both on the network instance and the app instance, which seems overkill.
Only on the network instance means that anybody and everybody can discover, which seems to relaxed.
But only on the app instance should be sufficient for the initial use for the sidecar container, and less to configure.

uncleDecart:
>>> I thought the idea was to only have setting on the app instance to allow discover. Am I confused?

>> We can move it to app instance level, but this way we can give access granularly, i.e. expose specific port of app instance

> The initial proposal was to have it both on the network instance and the app instance, which seems overkill.
Only on the network instance means that anybody and everybody can discover, which seems to relaxed.
But only on the app instance should be sufficient for the initial use for the sidecar container, and less to configure.

Makes sense, updated it with app instance. We should publish doc on lfedge wiki

proto/config/appconfig.proto
```diff
@@ -206,6 +206,10 @@ message AppInstanceConfig {
 
   // PatchEnvelopeRef is used to track all patch envelopes used by the app instance
   PatchEnvelopeRef patchRef = 23;
+
+  // allow to be discovered by other app instances in the same network
+  // instance. Default is false
```
eriknordmark, line 211: Comment is reversed; we want to grant the ability for an app instance to discover all the other app instances attached to its network instances. 
Unless @gkodali-zededa tells me otherwise.

gkodali-zededa: Yes. @eriknordmark is right.

eriknordmark (changes requested):
Please fix comment. Then ready to go.

uncleDecart:
> Please fix comment. Then ready to go.

@eriknordmark done

eriknordmark (approved):
LGTM

PR #50, opened 2024-03-21
Add flag to enable VNC for shim VM

europaul:
We differentiate between the access to the edge application and the
shim VM:
- enableVNC is used to enable VNC for the edge application
- enableVNCshimVM is used to enable VNC for the shim VM

This differentiation is need, because currently shim VM doesn't support
any user authentication, so anybody able to access the VNC port can
access the shim VM. This way VNC access to the shim VM stays disabled
by default unless explicitly enabled by the controller.


europaul:
@eriknordmark yetus compalins because `enableVncShimVM` is not in snake_case format, but I think we should leave it like that so it looks more similar to `enableVnc` flag, which is also not in snake_case

europaul:
>  If you want you can actually change enableVnc to enable_vnc (the generated golang header file ends up with the same CamelCase).

@eriknordmark I did a little research: changing the name from camelCase to snake_case like so
```
	EnableVnc  bool   `protobuf:"varint,16,opt,name=enableVnc,proto3" json:"enableVnc,omitempty"`
	EnableVnc  bool   `protobuf:"varint,16,opt,name=enable_vnc,json=enableVnc,proto3" json:"enable_vnc,omitempty"
```
might not break the compatibility with protobuf, but will break the compatibility with JSON (options at the end of the line).

eriknordmark:
> might not break the compatibility with protobuf, but will break the compatibility with JSON (options at the end of the line).

I looked around in the controller and I don't see us using the enableVnc in json. We do refer to the fields in the EVE API using their golang struct names (thus e,g., generation_count in info.proto is referred to as generationCount in the controller).

So there shouldn't be a problem changing this.

eriknordmark (approved):
LGTM but feel free to rename enableVnc as well.

europaul:
@eriknordmark yetus complains about the name change due to this [rule](https://buf.build/docs/breaking/rules#field_same_name). But if JSON names are not used anywhere we can probably ignore this.

eriknordmark:
> @eriknordmark yetus complains about the name change due to this [rule](https://buf.build/docs/breaking/rules#field_same_name). But if JSON names are not used anywhere we can probably ignore this.

Yes, we can ignore that in this case.

But can you squash to two commits? (one for proto files and one for the rest)?

PR #51, opened 2024-04-04
 Report SIM card type

milan-zededa:
This will allow to distinguish between a physical SIM card and an eSIM.
Older EVE versions will report the SIM card type as UNSPECIFIED.

Note that we will get this info from ModemManager:
https://www.freedesktop.org/software/ModemManager/doc/latest/ModemManager/ModemManager-Flags-and-Enumerations.html#MMSimType


eriknordmark (approved):
LGTM

PR #52, opened 2024-04-16
EVE Discovery service: move `allow_to_discover` to AppInst port 

uncleDecart:
This is more granular approach. From appInstance perspective,
you also know which network port will be used.


uncleDecart:
I see yetus [error](https://github.com/lf-edge/eve-api/actions/runs/8701137540/job/23862550960?pr=52#step:4:260) 

```
bufcompat:Previously present field "24" with name "allow_to_discover" on message "AppInstanceConfig" was deleted.
```

But I marked field 24 as reserved, not sure we should leave it there

proto/config/appconfig.proto
```diff
@@ -103,6 +103,9 @@ message SnapshotConfig {
 // can also be increased to cause an application instance restart without
 // any other change to the application instance.
 message AppInstanceConfig {
+  // deprecated fields
+  reserved 14, 24;
```
eriknordmark, line 107: How about also removing the comment with   // Deprecated 14

proto/config/netconfig.proto
```diff
@@ -55,6 +55,10 @@ message NetworkAdapter {
   // valid vlan id range: 2 - 4093
   // vlan id 1 is implicitly used by linux bridges
   uint32 access_vlan_id = 41;
+
+  // allow AppInstance to discover other AppInstances
+  // attached to its network instances. default is false
```
eriknordmark, line 60: Shouldn't this say "to this network instance" in singular?
Or to the network instance attached to this adapter?

eriknordmark (approved):
LGTM but some comment nits

proto/config/netconfig.proto
uncleDecart: renamed to "this network instance"

milan-zededa (approved):

PR #53, opened 2024-04-19
Allow to configure custom network adapter labels

milan-zededa:
EVE already offers adapter labels `uplink` and `freeuplink` to match
a specific group of ports to provide external connectivity for a network
instance.

We should allow users to define and assign their own adapter labels and
support the following new network instance use-cases:

1. Switch network instance with multiple ports: user should be able to
   select a group of ports, all of which will be bridged together under
   one switch NI. EVE should then automatically run the STP protocol for
   the bridge to avoid bridge loops and broadcast storms.
2. Local network instance with multiple ports: the routing table on the
   host (EVE) side will contain routes from all selected ports and will
   be used to select output port for the given flow.
   For (not only default) route we should:
     a) allow the user to pick the output port(s). Can be more than one
         for multipath routes.
     b) periodically test port connectivity and failover between ports
         when currently used one has broken connectivity
     c) allow to enable cost-based port selection
     d) allow to enable cellular signal strength based selection.
        Applicable only among wwan ports.
   _a_ and _b_/_c_/_d_ could be combined together to allow the user to select a
   subset of NI ports for a route and then let EVE to pick one of them
   based on the probing results and other selection criteria.
   We should also allow to limit a port-forwarding rule to only a subset
   of ports used by a local network instance. This can be done by
   defining a new `ACEMatch` type `adapter` where value is adapter label
   to match.

Currently, EVE supports a variant of the second use-case for hard-coded
`uplink` and `freeuplink` labels. It would make sense to unify the behaviour
between these existing labels and the user-defined ones. This means that
we would make small change to the current implementation:
when `uplink` or `freeuplink` is selected for a local network instance,
the NI routing table will have IP routes from *all* matching (mgmt)
ports and the default route will be selected by connectivity testing.
This is different to the present behaviour, where EVE installs only
routes of the port selected by connectivity testing into the NI routing
table. But lets say that there are 2 mgmt ports: eth0 and eth1, and
connectivity testing selected eth0. Then even if app is accessing hosts
which are directly inside the eth1 subnet, traffic will be routed via eth0,
which can be confusing. Also, user-defined static routes for eth1 are
not present in the routing table and therefore inactive when eth0 is
selected. The idea is to apply connectivity testing to only default
route selection.

Another change in behaviour is wrt. the inbound (port forwarding) rules.
Without the newly introduced `adapter` label specified (e.g. when EVE is
upgraded but config remains the same), the inbound rules will be applied
to all the ports used by a network instance. With `uplink` labels this
means that port forwarding will be enabled on all the management ports,
not just on the one with the default route activated. However, this new
behaviour is actually more practical for the user. Accessing applications
will be possible through any of the mgmt ports, without having to know
which port is being used by the NI for the default route. After all,
NI with the `uplink` label is configured to use whichever mgmt port is
working, meaning that the user should ensure that opening ports on any
of them is safe in their environment. It therefore makes no difference
if port forwarding rules are always activated on all the managment ports
vs. only on one of them.

Lastly, to allow cost-based port selection for routes with multiple
output ports, some of which are possibly app-shared and not enabled
for management, we should allow the controller to set cost to also
app-shared ports (this change only affects the controller, EVE API
already does not limit cost attribute to only mgmt ports).


eriknordmark:
@milan-zededa I think it makes sense generalizing what we have. However, for the connectivity testing we currently at least partially depend on them being management ports - we check connectivity to the controller. We also check connectivity to the default router. I wonder whether that default router check would be sufficient for non-manegement ports. (If not we have to expose the ability to configure probe target IPs and/or URLs to the controller, UI, etc.)

proto/config/fw.proto
```diff
@@ -10,6 +10,46 @@ option java_package = "org.lfedge.eve.config";
...
+  //              lport refers to 2222). lport can be combined with any other match type.
+  //              It is actually required to combine "lport" and "protocol" inside the same ACE.
+  //              In other words, port without protocol is not valid.
+  //   * "fport": value should be a remote endpoint port number (foreign port). Used for filtering
+  //              actions, but not for PORTMAP (do not confuse with "lport", which is still used to
+  //              represent the forwarded port - the forwarded port is still considered as local).
+  //              "fport" can be combined with any other match type. It is actually required
+  //              to combine "fport" with "protocol" inside the same ACE. In other words, port
+  //              without protocol is not valid.
+  //  * "adapter": value should be an adapter label (SystemAdapter.Labels). It can be used for
```
eriknordmark, line 47: Your example used "eth0" i.e., an interface name (or logical label?).
Assuming we want that capability we can document that a port always has an implicit label which is its (logical label, ifname, pick one?).

proto/config/netinst.proto
```diff
@@ -97,14 +116,29 @@ message NetworkInstanceConfig {
...
 
-  // port - Only a single port is supported.
-  //    This is used as the external connection for the network instance.
-  //    This can be a physical (eth0 ) or logical port (vlan 0).
-  //    The port name comes from DeviceConfig ( When it is supported in future).
-  //    If the user needs multiple physical ports, Device config should be
-  //    used to create a label for multiple physical ports.
+  // Port(s) providing external connectivity for the network instance.
+  // Either a single port referenced by its name (SystemAdapter.Name) or an adapter label
+  // matching multiple network ports (ports containing this label in SystemAdapter.Labels).
```
eriknordmark, line 121: The label can match zero or more ports - matching one port is quite common e.g., for "uplink" when there is only one management port.

proto/config/netinst.proto
```diff
@@ -97,14 +116,29 @@ message NetworkInstanceConfig {
...
+  // Note that apart from the user-defined adapter labels, EVE automatically assigns "all"
+  // label to every port, "uplink" label to every management port and "freeuplink" label
+  // to every management port with zero cost.
+  // Both physical (e.g. eth0) and logical (e.g. vlan0) ports are allowed.
+  // Network instance with no assigned port (empty port reference or label not matching
+  // any port) is air-gapped, i.e. completely isolated from external networks.
+  // With multiple ports assigned, the behaviour depends on the network instance type.
+  // Switch network instance will run Spanning Tree Protocol (STP) to avoid bridge loops
```
eriknordmark, line 129: Don't we need STP plus BPDU guard? Or will STP catch the common loops?
( I know from experience that STP doesn't catch loops if a port or cable is defective and only passes traffic in one direction, but that is an uncommon case.)

proto/config/fw.proto
```diff
@@ -10,6 +10,46 @@ option java_package = "org.lfedge.eve.config";
 
 message ACEMatch {
   // FIXME: We should convert this to enum
```
eriknordmark, line 12: Feel free to remove this FIXME. I don't think it makes sense to replace the strings with enums any more.

milan-zededa: I would actually prefer enum, but zedcloud would have to keep setting the `string` for older EVE versions anyway, so it is not really worth making the change.
FIXME comment removed.

proto/config/netinst.proto
```diff
@@ -85,6 +85,25 @@ message IPRoute {
...
+  // EVE will use the default option, which is connectivity testing for "uplink"
+  // and "freeuplink" port labels, and cost-based selection otherwise.
+  DEFAULT_ROUTE_SELECTION_UNSPECIFIED = 0;
+  // Periodically run connectivity tests against the port gateway IP (if available)
+  // and the controller URL (if management port), to apply default route from
+  // a port with working connectivity.
+  // This selection algorithm thus provides automatic fail-over between ports.
+  DEFAULT_ROUTE_SELECTION_CONNECTIVITY_TESTING = 1;
+  // Apply default route of the port with the lowest cost.
+  // This is limited to management ports because app-shared ports do not have cost defined.
```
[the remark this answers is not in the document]
milan-zededa: I'm not sure. Either we add cost to app-shared ports or this will be renamed to `DEFAULT_ROUTE_SELECTION_USER` and there will be another field in `NetworkInstanceConfig` to select the port to use by default (by name or label). The latter requires that the controller supports run-time change of the network instance config (which EVE does support).

proto/config/netinst.proto
milan-zededa: Comment updated.

proto/config/fw.proto
milan-zededa: Since application config (and fw rules) can be in theory reused for multiple edge devices (like the commercial controller does), the `adapter` value must be label and cannot reference interface name.
In my example I have label `uplink`: https://github.com/milan-zededa/eve-api/blob/adapter-labels/proto/config/fw.proto#L81
I think you are referencing the example for the network instance port, where the interface name is allowed: https://github.com/milan-zededa/eve-api/blob/adapter-labels/proto/config/netinst.proto#L125

proto/config/netinst.proto
milan-zededa: It probably make sense to automatically enable BPDU guard on every application VIF connected to the switch NI. These are effectively leaf (edge) ports and we do not expect them to participate in STP. It would also make sense if these ports were immediately set to the forwarding state. However, I do not see feature similar to CISCO's proprietary "Fast-Enabled port" being available on the Linux bridge.
Perhaps it also make sense to allow the user to select a subset of physical ports which are also expected to be leafs, or where the downstream bridges are not fully trusted. This can also be a string attribute expecting a port label (selecting zero or more ports). But in that case we will also have to allow to user to re-enable ports that were previously blocked. Due to the declarative nature of the config, this would have to be a counter attribute.

milan-zededa:
> @milan-zededa I think it makes sense generalizing what we have. However, for the connectivity testing we currently at least partially depend on them being management ports - we check connectivity to the controller. We also check connectivity to the default router. I wonder whether that default router check would be sufficient for non-manegement ports. (If not we have to expose the ability to configure probe target IPs and/or URLs to the controller, UI, etc.)

True, I will think about how to allow configuring probing endpoint for any port, and ideally also how to unify it with the cellular connectivity testing (https://github.com/milan-zededa/eve-api/blob/adapter-labels/proto/config/netconfig.proto#L91-L111),

naiming-zededa:
How will the static routing implementation be affected this change?

milan-zededa:
> How will the static routing implementation be affected this change?

Currently, NI routing table contains only routes corresponding to the single selected port (selected at the given moment by connectivity testing). This also applies to static IP routes, i.e. we do not install static routes that at the given moment do not point to the selected port.
What I would like to change, is that the port connectivity testing would only apply to the default route selection. Link-local and (non-default) static IP routes of all ports matched by NI label would always be present in the NI routing table.
@naiming-zededa Do you think this may break user expectation or not be suitable for some scenarios?

naiming-zededa:
@milan-zededa I think this is fine. Only that all the ip routes can be applied onto various interfaces with gateway address matching their interface prefixes, and default route most likely will be propagated since there is mix of mgmt and app-shared, all the connected routes can be propagated if the knob is selected.

uncleDecart (approved):
LGTM

proto/config/fw.proto
eriknordmark: L125 is no longer correct but it's been months. Were you referring to https://github.com/milan-zededa/eve-api/blob/0ee94e298044c9e7299db0b30d64d161887d1016/proto/config/netinst.proto#L176?

eriknordmark (commented):
One additional observation is that for a switch NI with multiple external ports, and when one or more of those ports are management ports, we are creating a coupling between the device management side of designating and using management ports, and the workload configuration (a switch NI for some workloads). The deployment of such a workload can potentially have a negative implact on the device connectivity.

An example is if eth0 is a management port and eth1 is app shared and a switch NI is configred to use eth0 and eth1 as external ports. If that NI is incorrect, then potentially an external Ethernet switch on eth0 might have its BPDU guard trigger resulting in disconnecting the management connectivity for the device.

I don't know if there  are additional  restrictions we should document in the API for the switch NI with multiple external ports, or work on safety mechanisms in the implementation.

Thouhts?

proto/config/fw.proto
```diff
@@ -9,7 +9,46 @@ option java_package = "org.lfedge.eve.config";
...
+  //   * "fport": value should be a remote endpoint port number (foreign port). Used for filtering
+  //              actions, but not for PORTMAP (do not confuse with "lport", which is still used to
+  //              represent the forwarded port - the forwarded port is still considered as local).
+  //              "fport" can be combined with any other match type. It is actually required
+  //              to combine "fport" with "protocol" inside the same ACE. In other words, port
+  //              without protocol is not valid.
+  //  * "adapter": value should be an adapter shared label (SystemAdapter.Sharedlabels).
+  //               It can be used for an inbound ACE to apply the rule only to packets arriving
+  //               via one of the matched network adapters. Typically used to activate a given
+  //               port-forwarding rule (PORTMAP) for only a subset of network adapters.
```
eriknordmark, line 49: Do we need to say what happens when this is not set for a port-forwarding rule? (I think it be applies as "uplink" in that case.)

proto/config/netinst.proto
```diff
@@ -82,7 +82,76 @@ message IPRoute {
...
+  // to the network instance.
   string gateway = 2;
+  // Output device port for the routed traffic.
+  // Either a single NI port referenced by its name (SystemAdapter.Name, aka logical label)
+  // or an adapter shared-label matching zero or more NI ports (multipath routing).
+  // Can be used additionally to or instead of `gateway` to determine the next hop
+  // and the output network interface for the destination network.
+  // At least one of 'gateway' or 'port' must be defined.
+  // Since 'port' refers to one or more physical network adapters, it is not applicable
+  // to routes where gateway points to another application running on EVE.
```
eriknordmark, line 96: s/application/application instance/

proto/info/info.proto
```diff
@@ -1068,8 +1082,37 @@ message ZInfoNetworkInstance {
...
+  string gateway = 2;
+  // Logical label of the output device port for the routed traffic.
+  // Empty if the gateway is IP address of one of the applications. In that case,
+  // 'gateway_app' is defined instead.
+  string port = 3;
+  // UUID of the application used as the gateway for the route.
+  // Empty if the gateway is external (not one of the apps but outside of the device)
+  // or this is a connected route for a port (thus gateway is empty).
+  // In that case, 'port' is defined instead.
+  string gateway_app = 4;
```
eriknordmark, line 1115: I'm reading the separate comments and it seems like they are saying that exactly one of gateway, port, and gateway_app is set. Or maybe gateway+port can both be set (useful if gateway is ever an IPv6 link-local).

It would be easier to understand the choices if there was a single comment explaining how the next hop can be specified instead of the per-field comments.

eriknordmark (commented):
Some clarifications and questions about the comments.

proto/config/netcmn.proto
```diff
@@ -114,3 +114,29 @@ enum WiFiKeyScheme {
...
+  // Method to use to determine the connectivity status.
+  ConnectivityProbeMethod probe_method = 1;
+  // IP or FQDN to probe using the selected probing mechanism to determine
+  // the connectivity status.
+  // For CONNECTIVITY_PROBE_METHOD_TCP the address should include port number
+  // in one of the formats:
+  //   ipv4:port
+  //   [ipv6]:port
+  //   hostname:port
+  string probe_address = 2;
```
naiming-zededa, line 141: what about 'probe_endpoint' which is address plus port at least for TCP type

milan-zededa:
> I don't know if there are additional restrictions we should document in the API for the switch NI with multiple external ports, or work on safety mechanisms in the implementation.
> Thouhts?

The priority right now is the Local NI with multiple ports. Implementation for switch NI with multiple ports will be done later. As a result, I haven't though much about it yet. We will have to impose some limitations and document them in the API once we figure out the best solution. So for now I just put there minimum information on the topic of switch NI with multiple ports to lay some groundwork, but later we will have to figure out all the details.
I added this comment there for now:
```
  // However, EVE may filter out some of the matched ports which are not compatible
  // with the network instance type (e.g. wireless adapters cannot be used with switch NI
  // and ports without IP addresses are not usable for local NIs).
```
(maybe we will have to exclude mgmt ports as well...)

proto/config/fw.proto
```diff
@@ -9,7 +9,46 @@ option java_package = "org.lfedge.eve.config";
...
+  //              lport refers to 2222). lport can be combined with any other match type.
+  //              It is actually required to combine "lport" and "protocol" inside the same ACE.
+  //              In other words, port without protocol is not valid.
+  //   * "fport": value should be a remote endpoint port number (foreign port). Used for filtering
+  //              actions, but not for PORTMAP (do not confuse with "lport", which is still used to
+  //              represent the forwarded port - the forwarded port is still considered as local).
+  //              "fport" can be combined with any other match type. It is actually required
+  //              to combine "fport" with "protocol" inside the same ACE. In other words, port
+  //              without protocol is not valid.
+  //  * "adapter": value should be an adapter shared label (SystemAdapter.Sharedlabels).
```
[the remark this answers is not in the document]
milan-zededa: Fixed

proto/config/fw.proto
milan-zededa: Added comment about the default behavior. It is actually equivalent to using the "all" label.

proto/config/netinst.proto
```diff
@@ -82,7 +82,76 @@ message IPRoute {
   string destination_network = 1;
   // Gateway IP address.
   // This must be a valid IP address and can not be all-zeroes.
+  // At least one of 'gateway' or 'port' (see below) must be defined (both can be set).
+  // This can be an external gateway IP or an IP address of one of the apps connected
```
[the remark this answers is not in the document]
milan-zededa: Done

proto/config/netinst.proto
milan-zededa: Done

proto/info/info.proto
milan-zededa: Added comment about next-hop for `message IPRoute` for clarification:
```
// The next-hop for the IP route is specified by the "gateway" IP address
// (empty for a connected route) and the output "port". However, if the gateway
// is one of the applications running on the device, the port is actually a virtual
// interface with an EVE-generated interface name and no user-assigned logical label.
// Therefore, in such cases, it is more useful to leave "port" empty and instead use
// the "gateway_app" field to publish the UUID of the application used as the gateway.
```

proto/config/netcmn.proto
milan-zededa: Done, added `message ProbeEndpoint`.

naiming-zededa (approved):
LGTM

proto/info/info.proto
```diff
@@ -1040,8 +1041,21 @@ message ZInfoNetworkInstance {
...
 
-  string CurrentUplinkIntf = 10;  // current uplink interface in use by this NI
-  string CurrentUplinkAlias = 11; // user defined namestring alias to this CurrentUplinkIntf, for future
+  // Current uplink interface in use by this NI.
+  // Deprecated - network instance can be associated with multiple device ports,
+  // not just one. To find out which ports are being used by network instance,
+  // read the reported field "ports" instead (see below). Additionally, "routes"
+  // informs about the routes configured for the NI and which ports are currently
+  // used to route the traffic out (this includes info for the default route).
+  string CurrentUplinkIntf = 10 [deprecated = true];
```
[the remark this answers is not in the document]
eriknordmark: I saw an example in https://stackoverflow.com/questions/60490487/whats-the-best-way-to-deprecate-a-field-in-protocol-buffer-v3-reserved-vs-depre but not a description of the syntax.

proto/info/info.proto
eriknordmark: That helps. However, it doesn't say whether gateway can/should be empty when its is an application running on the device, plus a reader will still try to parse comments for the individual fields  and their head will hurt just like mine.

How about also simplifying the comments for the fields to e.g.,
// Optional logical label of the output device for the route. Can not be used when the gateway is an app instance of the device.
string port ...
// Optional UUID of the app instance used as the gateway for the route.
string gateway_app ...

milan-zededa: But this is IP route definition for the `ZInfoNetworkInstance` info message (I think it makes sense to publish all routes configured for NI). So this is about what fields to expect, why sometimes they are empty, etc. It is not about what the user should set.

proto/info/info.proto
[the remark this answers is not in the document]
milan-zededa: This requires some new imports and protoc arguments which I would opt to avoid. I think that having the comment explaining the deprecation is good enough.

proto/info/info.proto
eriknordmark: Ok

eriknordmark (approved):
LGTM

PR #54, opened 2024-04-22
Fix allow to discover

uncleDecart:
PR https://github.com/lf-edge/eve-api/pull/46 accidently added allow_to_discover field to app_instance
which was removed in https://github.com/lf-edge/eve-api/pull/52

Not deprecating field number 25 (allow_to_discover) because we haven't
had released EVE with this API yet, so we don't have to do this


proto/config/appconfig.proto
```diff
@@ -210,10 +208,6 @@ message AppInstanceConfig {
   // PatchEnvelopeRef is used to track all patch envelopes used by the app instance
   PatchEnvelopeRef patchRef = 23;
 
-  // allow AppInstance to discover other AppInstances
```
eriknordmark, line 213: But this was already removed in #52 based on the diffs, right?

uncleDecart: Yes, but #46 introduced this field once again with different number, if you look at master [here](https://github.com/lf-edge/eve-api/blob/main/proto/config/appconfig.proto#L215), you'll see this field

eriknordmark: Ah - presumably due to some mismerge. Got it.

PR #56, opened 2024-04-30
Make MTU of network ports and network instances configurable

milan-zededa:
The main motivation is to allow EVE users to use Jumbo frames
(more than 1500 bytes of payload) to/from their applications.


proto/config/netconfig.proto
```diff
@@ -26,6 +26,17 @@ message NetworkConfig {
 
   // wireless specification
   WirelessConfig wireless = 10;
+
+  // Maximum transmission unit (MTU) to set on the associated network adapter(s).
+  // This is the largest IP packet that the underlying link can and is allowed to carry.
+  // This does not include the L2 header size (e.g. Ethernet header or a VLAN tag size).
+  // The value is a 16-byte unsigned integer, representing the MTU size in bytes.
+  // The minimum legal value for the MTU is 68 (RFC 791).
```
eriknordmark, line 34: Shouldn't we specify 1500 or 1460 as the min? 1460 is the min  MTU to carry IPv6. 576 I think is the old common min for IPv4.

milan-zededa: OK, but according to the IPv6 RFC, min MTU to carry IPv6 is 1280: https://datatracker.ietf.org/doc/html/rfc8200#section-5

proto/config/netinst.proto
```diff
@@ -153,4 +153,18 @@ message NetworkInstanceConfig {
...
+  // This is the largest IP packet that the network instance is allowed to carry.
+  // The value is a 16-byte unsigned integer, representing the MTU size in bytes.
+  // The minimum legal value for the MTU is 68 (RFC 791).
+  // If not defined (zero value), EVE will set the MTU to the default value of 1500 bytes.
+  // On the host side, MTU is set to interfaces by EVE. On the guest (application) side,
+  // it is the role of the DHCP client to apply the MTU value, which it will receive with
+  // the DHCP lease via option 26.
+  // Please note that application traffic leaving or entering the device via a network
+  // adapter associated with the network instance will be additionally limited by the MTU
+  // value of the adapter, configured inside the NetworkConfig object (see netconfig.proto).
```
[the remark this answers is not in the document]
milan-zededa: OK, makes sense, added comment about this.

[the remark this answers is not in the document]
milan-zededa: However, I'm not sure how to handle the switch network instance. In that case, the MTU for the network adapter and the MTU for the switch NI reference the same bridge. I believe we should then flag both the adapter and NI with an error if the MTUs are not the same (and use the lower of the two values for the time being). WDYT?

proto/config/netinst.proto
```diff
@@ -153,4 +153,21 @@ message NetworkInstanceConfig {
...
+  // Maximum transmission unit (MTU) to set for the network instance and all application
+  // interfaces connected to it.
+  // This is the largest IP packet that the network instance is allowed to carry.
+  // This does not include the L2 header size (e.g. Ethernet header or a VLAN tag size).
+  // The value is a 16-byte unsigned integer, representing the MTU size in bytes.
+  // The minimum accepted value for the MTU is 1280 (RFC 8200, "IPv6 minimum link MTU").
+  // If not defined (zero value), EVE will set the MTU to the default value of 1500 bytes.
+  // On the host side, MTU is set to interfaces by EVE. On the guest (application) side,
+  // it is the role of the DHCP client to apply the MTU value, which it will receive with
+  // the DHCP lease via option 26.
```
uncleDecart, line 166: what is option 26?

milan-zededa: https://datatracker.ietf.org/doc/html/rfc2132#section-5.1

milan-zededa: @eriknordmark @uncleDecart @naiming-zededa 
Speaking of MTU on the guest-side (inside the app), I wonder how to propagate MTU to applications connected to switch NIs (we do not run DHCP/SLAAC for those). For container apps run inside our shim-VM, we could write per-interface MTU to the kernel cmdline and apply them from our init-initrd script (e.g. "mtu=1500,1500,9000" for eth0, eth1and eth2, respectively). The only downside, is that while MTUs of VIFs connected to local NIs will be allowed to change in-runtime by use of DHCP lease renew (or IPv6 RA update), MTUs of switch VIFs would stay fixed after the shim-VM boot (probably acceptable limitation).
However, I do not have any idea how to automatically propagate MTU of switch NI VIFs to VM applications.

naiming-zededa: I would think pre-reboot setting should be good enough, we can say if the external network wants to change, then the external dhcp server is responsible to convey the mtu message.

milan-zededa: OK, that will do for **container** apps (where we initialize the shim-VM).
@naiming-zededa do you have some idea how we could automatically set MTUs of **VM** application interfaces connected to switch NIs? I believe this is completely out of our control, but maybe I'm missing something.

naiming-zededa: I'm not familiar with that, it is possible it can only be set through the cloud-init during the first bootup.

proto/config/netconfig.proto
eriknordmark: > OK, but according to the IPv6 RFC, min MTU to carry IPv6 is 1280: https://datatracker.ietf.org/doc/html/rfc8200#section-5

You're correct.

proto/config/netinst.proto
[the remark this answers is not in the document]
eriknordmark: Makes sense.

But what about a local network instance? Will the NAT fragment automatically if the external ports and internal bridge MTU differs? Or do we need to also check those for equality?

proto/config/netinst.proto
eriknordmark: Is it correct that the issue appears for a VM (Linux, Windows, whatever) using a switch network instance?
I wonder if for those we should explore whether we can detect from the host whether the vnic in the guest has a different MTU setting. Does that get carries across somehow?

milan-zededa: > Is it correct that the issue appears for a VM (Linux, Windows, whatever) using a switch network instance?
Yes, correct.
> I wonder if for those we should explore whether we can detect from the host whether the vnic in the guest has a different MTU setting. Does that get carries across somehow?
I'm not aware of anything, will investigate...

eriknordmark: For now we can just document that it is up to the user to make sure that the VMs are configured with jumbo-frames in addition to the network.

proto/config/netinst.proto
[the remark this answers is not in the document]
milan-zededa: I decided that we will be forcing MTU equality between NI and the associated adapter for both switch and local NIs. Otherwise we may have undesired IP fragmentation happening inside EVE.

proto/config/netinst.proto
milan-zededa: Updated documentation to explain all these cases and how user should deal with them.

milan-zededa:
@eriknordmark PR is ready for another round of review and merge.

eriknordmark (approved):
LGTM

PR #57, opened 2024-05-20
Report MTU used by network instance 

milan-zededa:
Now that network instance MTU is configurable, we should report what
the actual value being used is. This can differ from the configured
value in case it is invalid or conflicts with the MTU of the associated
device port.
Please note that we already report MTU of every device port.


naiming-zededa (approved):
LGTM

eriknordmark (approved):
LGTM

PR #58, opened 2024-05-22
 Clarification on the MTU propagation from host to guest 

milan-zededa:
Virtio interfaces do actually allow to propagate MTU value from the host to the guest.

### DIFF
--- a/go/config/netinst.pb.go
+++ b/go/config/netinst.pb.go
@@ -607,7 +607,8 @@ type NetworkInstanceConfig struct {
 	// If not defined (zero value), EVE will set the MTU to the default value of 1500 bytes.
 	// On the host side, MTU is set to interfaces by EVE. On the guest (app) side,
 	// the responsibility to set the MTU lies either with EVE or with the user/app, depending on
-	// the network instance and app type (local or switch, VM or container).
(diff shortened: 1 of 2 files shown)


eriknordmark (approved):
LGTM.
Is there some place where folks can determine whether VIRTIO_ NET_F_MTU is supported? ethtool? Something else?

milan-zededa:
> LGTM. Is there some place where folks can determine whether VIRTIO_ NET_F_MTU is supported? ethtool? Something else?

http://blog.vmsplice.net/2020/05/how-to-check-virtio-feature-bits-inside.html

Given that
```
#define VIRTIO_NET_F_MTU 3
```
We can check with:
```
# position argument of "cat" starts with 1, hence we have to do +1
cat /sys/class/net/enp1s0/device/features | cut -c 4
```

Maybe I can link this article from the comment?

milan-zededa:
@eriknordmark I will be submitting detailed documentation on the MTU topic for the EVE repo and will describe how to check VIRTIO_NET_F_MTU and other things.
I think that I would prefer to avoid making the comment any longer here with some details and adding links that may go broke. Users do not really read comments in this repo anyway but look into the EVE documentation. WDYT?

eriknordmark:
Ok

PR #59, opened 2024-06-06
Introduce new ZInfoNTPSource message in `proto/info/ntpsource.proto`

rouming:
`ZInfoNTPSource` message contains a list of peers describing to which NTP daemon (e.g. chronyd) is synchronized on EVE side. This message will be sent periodically from EVE to the `/info` controller endpoint.

CC: @zed-rishabh 
CC: @eriknordmark


rouming:
There is one yetus complain 
```
buflint:Enum zero value name "NTP_PEER_STATE_SYNC" should be suffixed with "_UNSPECIFIED".
```
which I can't fix, because 0 state means "SYNC" and not UNSPECIFIED.

proto/info/ntpsources.proto
```diff
@@ -0,0 +1,74 @@
...
+import "google/protobuf/timestamp.proto";
+
+package org.lfedge.eve.info;
+option go_package  = "github.com/lf-edge/eve-api/go/info";
+
+option java_package = "org.lfedge.eve.info";
+
+// Enum for describing the NTP source state
+enum NTPSourceState {
+  NTP_SOURCE_STATE_SYNC        = 0;
```
eriknordmark, line 15: This buflint should be fixed since zero needs to be reserved (by definiting an _UNSPECIFIED name) and the real values starting at 1.

proto/info/ntpsources.proto
```diff
@@ -0,0 +1,74 @@
...
+  NTP_SOURCE_STATE_UNREACH     = 1;
+  NTP_SOURCE_STATE_FALSETICKER = 2;
+  NTP_SOURCE_STATE_JITTERY     = 3;
+  NTP_SOURCE_STATE_CANDIDATE   = 4;
+  NTP_SOURCE_STATE_OUTLIER     = 5;
+}
+
+// Enum for describing the NTP source mode
+enum NTPSourceMode {
+  NTP_SOURCE_MODE_CLIENT   = 0;
```
eriknordmark, line 25: Ditto

eriknordmark:
> There is one yetus complain
> 
> ```
> buflint:Enum zero value name "NTP_PEER_STATE_SYNC" should be suffixed with "_UNSPECIFIED".
> ```
> 
> which I can't fix, because 0 state means "SYNC" and not UNSPECIFIED.

Is NTPSourceState and NTPSourceMode from some external/standard document? If so it would be good to put a reference to such a document.
And add whatever comment buflint needs to disable the check for that line. See https://buf.build/docs/configuration/v1/buf-yaml#allow_comment_ignores

rouming:
> > which I can't fix, because 0 state means "SYNC" and not UNSPECIFIED.
> 
> Is NTPSourceState and NTPSourceMode from some external/standard document? If so it would be good to put a reference to such a document. And add whatever comment buflint needs to disable the check for that line. 

mode and states are from the chrony itself. I added explanation and link to the sources.

I fixed the yetus for _UNSPECIFIED part by keeping state and mode +1, but there are more not related to that. I do not want to spoil each line with some yetus-disabling comment. This makes code looks ugly and the overall impression is that not a tool for development, but development for the tool making it happy. As usual 90% useless.

rouming:
@eriknordmark could you please take a look and approve/merge if no objections. Want this changes to be taken by the other teams. Thanks.

eriknordmark (approved):
LGTM

proto/info/ntpsources.proto
```diff
@@ -0,0 +1,80 @@
...
+  google.protobuf.Timestamp ref_time = 15;
+  sint32   poll                      = 16;
+  uint32   flags                     = 17; // bitmap from NTPSourceBitmap
+  double   offset                    = 18;
+  double   delay                     = 19;
+  double   dispersion                = 20;
+  double   jitter                    = 21;
+  double   root_delay                = 22;
+  double   root_disp                 = 23;
+}
```
zed-rishabh, line 76: What will be the unique identifier?

PR #60, opened 2024-06-08
add clusterToken to the encryption block message for cluster token

naiming-zededa:
- for message 'EdgeNodeCluster', we need encrypted string to be defined in cyber block


eriknordmark (approved):
LGTM

PR #61, opened 2024-07-01
Add replicated type volume

zedi-pramodh:
This commit introduces a volume type replicated.  Controller can set that type when sending app config for replicated nodes in a cluster.  Volumemgr in eve can use that flag to process these volumes differently ie wait for replica to show up etc.


deitch:
Ignoring the irrelevant yetus complaint.

deitch:
Is this ready to go @zedi-pramodh ?

zedi-pramodh:
I guess so, unless Erik or Naiming has questions/concerns

proto/config/storage.proto
```diff
@@ -171,8 +171,9 @@ enum VolumeAccessProtocols {
 
 enum VolumeContentOriginType {
   VCOT_UNKNOWN = 0;
-  VCOT_BLANK = 1;     // Create an empty directory/disk with specified size
-  VCOT_DOWNLOAD = 2;  // Create from downloadContentTreeID
+  VCOT_BLANK = 1;       // Create an empty directory/disk with specified size
+  VCOT_DOWNLOAD = 2;    // Create from downloadContentTreeID
+  VCOT_REPLICATED = 3;  // Use replicated volume from other nodes
```
naiming-zededa, line 176: but if it's 'replicated' flag set, do we have no need for 'download' or 'blank' flags anymore?

zedi-pramodh: Yes for replicated volumes. But there will be non-replicated volumes or this node could be master of a replicated volume so that will still have download or blank

eriknordmark: But do we want to support blank replicated volumes e.g., for a data volume which is initially empty?

Or is the intent that one of the nodes receives the config with DOWNLOAD or BLANK, and the other nodes in the cluster receives the config with thy type set to REPLICATED?

zedi-pramodh: > Or is the intent that one of the nodes receives the config with DOWNLOAD or BLANK, and the other nodes in the cluster receives the config with thy type set to REPLICATED?

Yes this is the intent. A node creating the volume will get BLANK or DOWNLOADED, other nodes will get REPLICATED type in the config

eriknordmark: But that approach does not enable an autonomous cluster during the creation (and I suspect it would be complex with race condition to have the controller assist with recovery should the creating node fail and either come back alive or not).

So I think the right approach here as well as elsewhere in the API is to tell all the nodes in the cluster the intended state (which in this case is that a volume should exist across the cluster) and if need he add hits and suggestions from the controller (such as the device uuid if the suggested creator - sent to all the nodes).

With such an API approach the initial implementation can rely solely on the suggested creator, but later we can make the cluster elect to ignore that should that node die. Thus we can improve robustness and autonomy of the cluster without neither adding to the API nor making the controller do something to help in recovery.

My intention was to have all aspects of the cluster API additions be done following that approach. I don't know if we did that for the previous additions but it might make sense to take a look.

eriknordmark (changes requested):
Need to declaratively state the desired state to all members of the cluster.

eriknordmark:
Should this pr be closed?

zedi-pramodh:
yes I will close this PR for now and in future if we need to add anything for volume management  I create a new PR

PR #62, opened 2024-07-22
 Add option to enable/disable flow logging for network instance

milan-zededa:
Flow-logging is an EVE feature that allows to collect information
about every application network flow. This includes attributes like
src/dst IP, src/dst port, ACL rule applied, packet/byte counters, etc.

However, this feature is always enabled for every network instance and
can produce lot of data which are then uploaded to the cloud.
We have seen cases where this amounted to hundreds of GB each week,
which was a burden for the controller's database.

Another drawback of flow-logging is that the iptables rules that EVE
installs for network instances are considerably more complicated because
of this feature and thus introduce additional packet processing overhead.

The goal of this API addition is to make flow logging configurable per-NI
and preferably disabled by default.


proto/config/netinst.proto
```diff
@@ -190,4 +190,16 @@ message NetworkInstanceConfig {
...
+  // If enabled, EVE periodically records all application TCP and UDP flows.
+  // A flow record encapsulates the application UUID, VIF name, open/close timestamps,
+  // src/dst IP/port/proto 5-tuple, and packet and byte counters.
+  // Additionally, EVE records every DNS request made from an application. This includes
+  // the request time, the hostname being resolved, and the returned IP address(es).
+  // The Flow logging API can be found in proto/config/flowlog/flowlog.proto
+  // By default, this feature is disabled. This is because it may potentially produce
+  // a large amount of data, which is then uploaded to the controller, and depending on
+  // the implementation, it may also introduce additional packet processing overhead.
+  bool enable_flowlog = 50;
```
eriknordmark, line 204: This being an "enable" means that the default will be off since the controller(s) would not set it initially. I wonder if it is better making it a "disable" knob and then we can add a properly documented and notified default change to disable it in the controller.

naiming-zededa: Or we can offer a configitem for this, if user do want this, they can explicitly enable this per device

milan-zededa: OK, changed to `disable_flowlog`.
Originally, I intended to make this a global configitem, but Daniel asked if this could be a NI option because configitems are not UX friendly.

eriknordmark (approved):
LGTM

proto/config/netinst.proto
zed-rishabh: I don't see any difference between having an ```enableFlowlog``` variable set to False by default and a ```disableFlowlog``` variable set to True. Both approaches achieve the same result: disabling flow logs by default. The key point is that the default state of flow logs is determined by the initial value of the variable. Whether we use ```enableFlowlog = False``` or ```disableFlowlog = True```, the outcome remains the same.

PR #63, opened 2024-08-09
Edge-Node-Cluster: node, pod, volume, replica, upgrade health.

andrewd-zededa:
Edge Node Clusters: nodes, volumes, replicas, upgrades
    
Node Info: role, age, ip
Pod Info: status, restarts, ip
Volume and Replica Info: rebuild progress, health
Upgrade Status: cluster wide and node components


andrewd-zededa:
Noticed a bad option in proto/info/edge_node_cluster.proto, fixing that in the next push.

proto/info/edge_node_cluster.proto
```diff
@@ -0,0 +1,236 @@
...
+  string name = 1;
+
+  // Status of the node
+  KubeNodeStatus status = 2;
+
+  // Role of the node is server or not
+  bool roleServer = 3;
+
+  // Age of the node, in seconds ago
+  uint64 age = 4;
```
milan-zededa, line 32: Since there can be a delay between info message published and received by the controller, wouldn't an absolute timestamp be better?

proto/info/edge_node_cluster.proto
```diff
@@ -0,0 +1,236 @@
...
+
+  // Version of the API Server running on the node
+  string apiServerVersion = 5;
+
+  // Internal IP address of the node
+  string internalIP = 6;
```
milan-zededa, line 38: How about naming this clusterIP? "Internal" sounds like it is not used outside the node.

proto/info/edge_node_cluster.proto
```diff
@@ -0,0 +1,236 @@
...
+
+  // status of the current component under upgrade, KUBE_COMP_UPGRADE_STATUS_UNKNOWN when 
+  //   no upgrades in progress
+  KubeCompUpgradeStatus status = 3;
+ 
+  // error info in case of failure
+  ErrorInfo error = 4;
+}
+
+message KubeClusterInfo {
```
milan-zededa, line 221: Is this published by every node in the cluster? (I assume it would be a duplicate info)

proto/info/edge_node_cluster.proto
```diff
@@ -0,0 +1,236 @@
...
+  // Replicas of the volume
+  repeated KubeVolumeReplicaInfo replica = 8;
+}
+
+message KubeStorageInfo {
+  // Overall status of Longhorn
+  StorageHealthStatus health = 1;
+
+  // time of the health status transition, seconds ago
+  uint64 transitionTime = 2;
```
milan-zededa, line 181: Again I think that absolute timestamp would be better given that this is published to a remote controller.

proto/info/edge_node_cluster.proto
```diff
@@ -0,0 +1,236 @@
...
+  string name = 1;
+
+  // Status of the volume
+  StorageVolumeState state = 2;
+
+  // Robustness of the volume
+  StorageVolumeRobustness robustness = 3;
+
+  // Age of the volume, seconds ago
+  uint64 age = 4;
```
milan-zededa, line 161: +1 for absolute timestamp

proto/info/edge_node_cluster.proto
```diff
@@ -0,0 +1,236 @@
...
+  // Application Status
+  KubePodStatus status = 2;
+
+  // Restart count of the application
+  uint32 restartCount = 3;
+
+  // Restart time of the application, seconds ago
+  uint64 restartTime = 4;
+
+  // Age of the application, seconds ago
```
milan-zededa, line 129: +1 for absolute timestamp

proto/info/edge_node_cluster.proto
```diff
@@ -0,0 +1,236 @@
...
+  // Name of the EVE application
+  string name = 1;
+
+  // Application Status
+  KubePodStatus status = 2;
+
+  // Restart count of the application
+  uint32 restartCount = 3;
+
+  // Restart time of the application, seconds ago
```
milan-zededa, line 126: +1 for absolute timestamp

milan-zededa:
Yetus has some opinions on the naiming of enum values...

proto/info/edge_node_cluster.proto
milan-zededa: Never mind, now I see that "kubectl get nodes" calls it INTERNAL-IP ...

andrewd-zededa:
@milan-zededa I don't see a 'make yetus' rule like the eve repo has, is there another convenience method to kick it off locally with the rules for this repo?

proto/info/edge_node_cluster.proto
andrewd-zededa: Yeah I can change the timestamps to a "google.protobuf.Timestamp" type instead.

andrewd-zededa:
Addressed the type concerns for relative -> absolute times now using google.protobuf.Timestamp.  I think I'll push another change to rename the 'age' fields to 'creation_timestamp' to match field name in Kubernetes json output.

andrewd-zededa:
Latest commit renames 'age' fields to absolute style 'creation_timestamp'

proto/info/edge_node_cluster.proto
```diff
@@ -0,0 +1,237 @@
...
+  string api_server_sersion = 5;
+
+  // Internal IP address of the node
+  string internal_ip = 6;
+}
+
+enum StorageHealthStatus {
+  STORAGE_HEALTH_STATUS_UNKNOWN = 0;
+  STORAGE_HEALTH_STATUS_HEALTHY = 1;
+  STORAGE_HEALTH_STATUS_DEGRAEDED_2_REPLICA_AVAILABLE_REPLICATING = 2;
```
eriknordmark, line 45: typo?

proto/info/edge_node_cluster.proto
```diff
@@ -0,0 +1,237 @@
...
+  // Name of the EVE application
+  string name = 1;
+
+  // Application Status
+  KubePodStatus status = 2;
+
+  // Restart count of the application
+  uint32 restart_count = 3;
+
+  // Restart time of the application, seconds ago
```
eriknordmark, line 127: Is this comment correct that it is a relative time? If so it would make sense to make this explicit because we have some place where an uptime (in seconds) displays as a time in 1970.

proto/info/edge_node_cluster.proto
```diff
@@ -0,0 +1,237 @@
...
+  KUBE_NODE_STATUS_READY = 1;
+  KUBE_NODE_STATUS_NOT_READY = 2;
+  KUBE_NODE_STATUS_SCHEDULING_DISABLED = 3;
+  KUBE_NODE_STATUS_OUT_OF_DISK = 4;
+  KUBE_NODE_STATUS_MEMORY_PRESSURE = 5;
+}
+
+message KubeNodeInfo {
+  // Name of the node
+  string name = 1;
```
eriknordmark, line 24: Is the intent that this name, node_name below, and other names in EVE-OS come from which namespace?

FWIW EVE-OS has a hostname (once onboarded to the controller) which is the device UUID, and VM has a hostname (from DHCP I think) which is the app instance UUID. Will this KubeNodeInfo name be one of those?

And will the pod node_name match?

proto/info/edge_node_cluster.proto
```diff
@@ -0,0 +1,237 @@
...
+syntax = "proto3";
+
+package org.lfedge.eve.info;
+option go_package = "github.com/lf-edge/eve-api/go/info";
+option java_package = "org.lfedge.eve.info";
+
+import "google/protobuf/timestamp.proto";
+import "info/info.proto";
+
+enum KubeNodeStatus {
```
eriknordmark, line 13: I assume this and other enums come from some kubernetes/longhorn source. Would it make sense to put comments indicating the source? That might be helpful if additional values are added later.

proto/info/edge_node_cluster.proto
```diff
@@ -0,0 +1,237 @@
...
+
+  // Replicas of the volume
+  repeated KubeVolumeReplicaInfo replica = 8;
+}
+
+message KubeStorageInfo {
+  // Overall status of Longhorn
+  StorageHealthStatus health = 1;
+
+  // time of the health status transition
```
eriknordmark, line 181: Is this the most recent transition? If so, please say that in the comment.

proto/info/edge_node_cluster.proto
```diff
@@ -0,0 +1,237 @@
...
+  google.protobuf.Timestamp transition_time = 2;
+
+  // Status of all the volumes in Longhorn
+  repeated KubeVolumeInfo volumes = 3;
+}
+
+enum KubeCompUpgradeStatus {
```
eriknordmark, line 188: What does "comp" mean here and in the next enum?

proto/info/edge_node_cluster.proto
```diff
@@ -0,0 +1,237 @@
...
+  KUBE_COMP_UPGRADE_STATUS_IN_PROGRESS = 3;
+  KUBE_COMP_UPGRADE_STATUS_FAILED = 4;
+  KUBE_COMP_UPGRADE_STATUS_COMPLETED = 5;
+}
+
+enum KubeComp {
+  KUBE_COMP_UNKNOWN = 0;
+  KUBE_COMP_CONTAINERD = 1; //every node will publish
+  KUBE_COMP_K3S = 2; // every node will publish
+  KUBE_COMP_MULTUS = 3; // only the seed node will publish the remaining here and below
```
eriknordmark, line 201: Seed node might have died and been replaced months ago. So I think there might be an elected leader which will publish, and that leader would initially be the seed node, but the concept of a seed node is a transient designation for initial creation.

proto/info/edge_node_cluster.proto
andrewd-zededa: Yes, I can add them.

proto/info/edge_node_cluster.proto
andrewd-zededa: Yes I'll fix this.  The first cluster node to upgrade was originally going to be the one listed as seed node, I'll change this to the 'first node to upgrade' to a given eve-os version.

proto/info/edge_node_cluster.proto
andrewd-zededa: Comp is an installed component in the cluster after first boot completes for eve, some of these are installed once on each node (k3s/containerd) vs others which are installed once per cluster (multus, longhorn, ...).

proto/info/edge_node_cluster.proto
andrewd-zededa: This is published by one node chosen through leader election.

proto/info/edge_node_cluster.proto
andrewd-zededa: fixed comment

proto/info/edge_node_cluster.proto
andrewd-zededa: > Is the intent that this name, node_name below, and other names in EVE-OS come from which namespace?

This will be all namespaces.

> FWIW EVE-OS has a hostname (once onboarded to the controller) which is the device UUID, and VM has a hostname (from DHCP I think) which is the app instance UUID. Will this KubeNodeInfo name be one of those?

name would be the node name as shown in the controller to match EdgeNodeInfo DeviceName

> And will the pod node_name match?

Yes

proto/info/edge_node_cluster.proto
andrewd-zededa: Yes, fixed

andrewd-zededa:
Latest commits had to pull ErrorInfo out to a separate file to resolve dependency issues created when adding KubeClusterInfo to InfoContent.

eriknordmark:
Looks like there are still build issues

andrewd-zededa:
Addressed some yetus issues I missed, and separate commit for the generated go, python, svg files.

andrewd-zededa:
Reorganized and moved back ErrorInfo due to the yetus issues.  I have a yetus target in my local copy of the makefile which I've been experimenting with using to try to catch these issues before a PR but it seems its not catching enough yet.

eriknordmark (approved):
LGTM

andrewd-zededa:
The publishing process of the top level message KubeClusterUpgradeStatus can be simplified if it is allowed to be a top level message type to be published.  Only one node will be upgrading/updating at any time so no leader election should be needed to publish KubeClusterUpgradeStatus if the currently updating node is the only node which publishes it.  The message would only be sent while the edge-node reports that it is in upgrade-test.

@naiming-zededa What do you think?

naiming-zededa:
@andrewd-zededa makes sense, since the nodes are upgrade one at a time and it needs to have controller connection at the time, so update by it's own is fine.

andrewd-zededa:
Latest change elevates KubeClusterUpgradeStatus into the InfoContent field of ZInfoMsg.

proto/info/info.proto
```diff
@@ -1316,3 +1319,34 @@ message ZInfoLocation {
...
+  repeated KubePodNameSpaceInfo pod_name_spaces = 2;
+
+  // Info message on a list of EVE applications
+  repeated KubeEVEAppPodInfo eve_apps = 3;
+
+  // Info message on cluster storage
+  KubeStorageInfo storage = 4;
+
+  // The current status of the cluster upgrade
+  KubeClusterUpgradeStatus upgrade = 5;
```
naiming-zededa, line 1351: What will this 'upgrade' report as part of the KubeClusterInfo ?

andrewd-zededa: It would report unspecified or the last component updated but this is a good point, it might be appropriate to remove it from KubeClusterInfo.

andrewd-zededa: latest change removes upgrade from KubeClusterInfo

proto/info/edge_node_cluster.proto
eriknordmark: It would make sense to add comments before each enum and message which would spell this out; KubeCompUpgradeStatus enumerates thepossible  status for the kubernetes components, or something in that vein.

andrewd-zededa:
I'll fix a mismatch I see between 'upgrade' / 'update' conventions, I'll fix those.

eriknordmark:
@andrewd-zededa looks like you got a conflict with Pramodh's PR.
Are there some remaining issues causing yetus failures?

andrewd-zededa:
Yeah I'll fix the conflict, looks like I just need to regenerate the built files

andrewd-zededa:
Updated comments on KubePodStatus, rename upgrade->update to match existing baseos update.  Rebase on main and rebuilt generated files to resolve conflict.

andrewd-zededa:
Looks like the two info messages KubeClusterInfo and KubeClusterUpdateStatus should probably be renamed to ZInfoKubeClusterInfo and ZInfoKubeClusterUpdateStatus but these types are getting a bit long so I'm open to suggestions to shorten them.

andrewd-zededa:
Looks like the two info messages KubeClusterInfo and KubeClusterUpdateStatus should probably be renamed to ZInfoKubeClusterInfo and ZInfoKubeClusterUpdateStatus but these types are getting a bit long so I'm open to suggestions to shorten them.

andrewd-zededa:
Another commit pending to append to the ZInfoTypes enum.

PR #64, opened 2024-08-15
Allow to use switch NI ports as VLAN access ports

milan-zededa:
The main use-case of switch NI with multiple ports is not actually to
bridge multiple L2 switches for redundancy and run STP (*) to avoid loops,
but to allow connecting external end devices (e.g. sensors) directly
into the same L2 segment as the applications running on the edge node.
With VLAN-enabled Switch NI, the user should be able to use a physical
network port attached to Switch NI as a VLAN access port (by default
all switch NI ports are configured as trunks).

(*) Regardless of the use-case, we will always run STP on a multi-port
switch NI to avoid accidental bridge loops.


proto/config/netinst.proto
```diff
@@ -310,4 +310,21 @@ message NetworkInstanceConfig {
...
+  repeated VlanAccessPort vlan_access_ports = 51;
 }
+
+// VLAN Access Port config applied to physical port(s) attached to a Switch Network Instance.
+message VlanAccessPort {
+  // VLAN ID from the range <1,4094>
+  uint32 vlan_id = 1;
+  // Physical network adapter used as the access port for the given VLAN.
+  // Either a single NI port referenced by its name (SystemAdapter.Name, aka logical label)
+  // or an adapter shared-label matching zero or more NI ports.
```
eriknordmark, line 328: Does it make sense even when it matches zero ports? Is this so it will wait for matching ports to appear later?

milan-zededa: This is just a generic description for shared labels.
Config that never matches any port is indeed pointless. However, there can be a case where temporarily no port is matched but later a new port appears (that was previously e.g. assigned to an app) or user adds shared label to an existing port. Zedrouter will then add the port into the switch NI and apply the VLAN config.

eriknordmark: ok

eriknordmark (approved):
LGTM

PR #67, opened 2024-09-20
Edge-Node-Cluster cleanup KubeVolumeInfo and KubeStorageInfo

andrewd-zededa:
- StorageHealthStatus should be under KubeVolumeInfo as a substatus.
- New enum ServiceStatus more appropriately describes multi-node service health.
- Added missing enum entries to StorageVolumeReplicaStatus


andrewd-zededa:
Moved a field from KubeStorageInfo to KubeVolumeInfo but I'm hoping it should be ok as these should not be in use yet as they were just added recently in PR #63

andrewd-zededa:
latest push should resolve yetus enum prefix issue

eriknordmark (approved):
LGTM.
As you say, if there is not yet any use of KubeStorageInfo.health neither in EVE nor in any controller, then there isn't a problem doing this protobuf type change in. Did you check all of those?

andrewd-zededa:
@eriknordmark I believe that eve-api version has been pulled into a dev branch of a controller but I don't think it has been pushed to a release of a production facing controller yet.  I think @xyuria-zededa can help confirm.

xyuria-zededa:
> @eriknordmark I believe that eve-api version has been pulled into a dev branch of a controller but I don't think it has been pushed to a release of a production facing controller yet. I think @xyuria-zededa can help confirm.

I haven't used it yet, so you can feel free to change anything there.

PR #68, opened 2024-09-25
Edge-Node-Cluster: cluster admission additions

andrewd-zededa:
- ZInfoKubeCluster needs a cluster_id which should match the id in config.EdgeNodeCluster.cluster_id
    
- Track node entrance into or out of a cluster with KubeNodeInfo.admission_status (NodeAdmission)


andrewd-zededa:
@naiming-zededa After some discussion with @xyuria-zededa we felt there were two missing fields here.  admission_status will help the controller to show the eventual-consistent state of a node entering/leaving a cluster.  cluster_id necessary for a controller to link ZInfoKubeCluster to existing cluster info. 

If you two have a moment to review I would appreciate it, I don't see an option for me to add you two as reviewers here.

naiming-zededa (commented):
LGTM

proto/info/info.proto
```diff
@@ -1348,4 +1348,7 @@ message ZInfoKubeCluster {
 
   // Info message on cluster storage
   KubeStorageInfo storage = 4;
+
+  // Cluster uuid
+  string cluster_id = 5;
```
eriknordmark, line 1353: Is this the UUID assigned by the controller for the edge node cluster object, or is it some uuid generated by k3s? It makes sense to clarify in the comment what it is.

andrewd-zededa: this should be the same id passed to the edge-node in the config message EdgeNodeCluster, I'll update the comment.

eriknordmark (approved):
LGTM

PR #69, opened 2024-10-09
Edge-Node-Cluster VMI additions

andrewd-zededa:
Track VirtualMachineInstance for eve VMs.


andrewd-zededa:
@naiming-zededa @xyuria-zededa Please review as time allows.

naiming-zededa (approved):
LGTM

proto/info/edge_node_cluster.proto
```diff
@@ -269,3 +272,37 @@ enum KubeComp {
...
+  // Name of the VirtualMachineInstance
+  string name = 1;
+
+  // Status (VirtualMachineInstance.Status.Phase)
+  KubeVMIStatus status = 2;
+
+  // Creation Time of the VirtualMachineInstance
+  google.protobuf.Timestamp creation_time = 3;
+
+  // LastTransitionTime is the last time the Ready Condition was true
```
eriknordmark, line 300: Does "the Ready Condition" refer to the is_ready field below? If so it makes sense to update this comment.
And I'd say "was set to true" and not "was true".  Would the timestamp also be updated when it is set to false?

eriknordmark (approved):
LGTM apart from a comment question/comment.

proto/info/edge_node_cluster.proto
andrewd-zededa: I just fixed the comments the current comment does leave a hole.  The timestamp should be updated for both ->ready and ->not-ready transitions.

andrewd-zededa:
> LGTM apart from a comment question/comment.

Addressed it in the latest push.

PR #70, opened 2024-10-11
proto: enhance ntp configuration

christoph-zededa:
1. allow to set more than one NTP server
2. allow to make NTP servers from cloud controller exclusively used, i.e. the NTP servers that got announced via DHCP are not used (in certain cases this an attacker might send DHCP responses with a different NTP server set and therefore can control time on EVE)


proto/config/devconfig.proto
```diff
@@ -169,6 +169,14 @@ message EdgeDevConfig {
 
   // cluster configuration
   EdgeNodeCluster cluster = 43;
+  DhcpVendorExtensionsOverride dhcp_extensions_override = 44;
+}
+
+// DhcpVendorExtensionsOverride is used to override or add dhcp vendor extensions like f.e.
+// the NTP servers - currently no other options are supported, but DNS servers might be a
+// good candidate in the future
+message DhcpVendorExtensionsOverride {
```
milan-zededa, line 178: I think that terminology 'vendor extensions' is more used with BOOTP. For DHCP the [RFC](https://datatracker.ietf.org/doc/html/rfc2132#section-2) just refers to them as Options.
So what about calling this `DhcpOptionsOverride`?

proto/config/devconfig.proto
```diff
@@ -169,6 +169,14 @@ message EdgeDevConfig {
 
   // cluster configuration
   EdgeNodeCluster cluster = 43;
+  DhcpVendorExtensionsOverride dhcp_extensions_override = 44;
```
milan-zededa, line 172: I think this should be part of `NetworkConfig` instead. Either inside `ipspec` or probaly better as another field.

christoph-zededa: You can have several `NetworkConfig`s - what is the meaning if you set it in one to `true` and in another one to `false`?
That means that for network device eth0 you can override the dhcp option and for another network device eth1 you don't override the dhcp option.
Is my interpretation correct?

proto/config/devconfig.proto
```diff
@@ -169,6 +169,14 @@ message EdgeDevConfig {
...
+}
+
+// DhcpVendorExtensionsOverride is used to override or add dhcp vendor extensions like f.e.
+// the NTP servers - currently no other options are supported, but DNS servers might be a
+// good candidate in the future
+message DhcpVendorExtensionsOverride {
+  bool ntpServerExclusively = 1; // use exclusively specified NTP servers
```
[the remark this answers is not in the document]
christoph-zededa: the NTP servers come from here: https://github.com/lf-edge/eve-api/pull/70/files#diff-b6af1b4801076bb7a7c0b5f919ecab887a34b218f6a5f2bfdffa40845a5e618dR769-R770

Currently they are only used if there is a static network configuration - I want to change it, to even use it if the network configuration is done via dhcp.

proto/config/devconfig.proto
milan-zededa: Yes, that would be the preferred behavior.

proto/config/devconfig.proto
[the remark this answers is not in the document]
milan-zededa: That link points to info.proto, i.e. status reported by device to zedcloud (we need to report these additional NTP servers, so that is correct).
But we will need such field also in this override structure or in `ipspec` which is used by `NetworkConfig`.

proto/config/netcmn.proto
```diff
@@ -84,11 +84,21 @@ message ipspec {
   string gateway = 5;
   string domain = 6;
   string ntp = 7;
+  repeated string moreNtp = 10;
```
eriknordmark, line 87: How abot calling this more_ntp_servers and adding a comment that ntp+more_ntp_servers is used to specify a set of NTP servers. Also, can these be hostnames or IP addresses? Makes sense to state that in a comment.

proto/config/netcmn.proto
```diff
@@ -84,11 +84,21 @@ message ipspec {
...
   ipRange dhcpRange = 9;
+
+  DhcpOptionsOverride dhcp_options_override = 11;
+}
+
+// DhcpOptionsOverride is used to override or add dhcp options like f.e.
+// the NTP servers - currently no other options are supported, but DNS servers might be a
+// good candidate in the future
+message DhcpOptionsOverride {
+  bool ntpServerExclusively = 1; // use exclusively specified NTP servers
```
eriknordmark, line 101: For these new fields it makes sense to follow the buflint about lower_snake_case.

proto/config/netcmn.proto
```diff
@@ -84,11 +84,21 @@ message ipspec {
...
   repeated string dns = 8;
 
   // for IPAM management when dhcp is turned on.
   // If none provided, system will default pool.
   ipRange dhcpRange = 9;
+
+  DhcpOptionsOverride dhcp_options_override = 11;
+}
+
+// DhcpOptionsOverride is used to override or add dhcp options like f.e.
```
eriknordmark, line 97: Are the semantics to override or to ignore? Stated differently, if there are no ntp servers specified but ntp_server_exclusively is true, will the result be that no NTP server configured? Or will the ntp_server_exclusively be ignored if no NTP server is configured in the API?

Makes sense to be explicit and make the name match the intended semantics.

proto/config/devconfig.proto
christoph-zededa: I changed it to `DhcpOptionsIgnore` https://github.com/lf-edge/eve-api/pull/70/files#diff-35ea4ed74707f6ad6d17be169425a7acf4e25e09e9de81a0f7300ed80c5ec656R101

proto/config/netcmn.proto
christoph-zededa: I changed it to DhcpOptionsIgnore https://github.com/lf-edge/eve-api/pull/70/files#diff-35ea4ed74707f6ad6d17be169425a7acf4e25e09e9de81a0f7300ed80c5ec656R101

And I added a comment regarding what should happen if no ntp servers are specified: https://github.com/lf-edge/eve-api/pull/70/files#diff-35ea4ed74707f6ad6d17be169425a7acf4e25e09e9de81a0f7300ed80c5ec656R103

proto/config/netcmn.proto
christoph-zededa: Done - https://github.com/lf-edge/eve-api/pull/70/files#diff-35ea4ed74707f6ad6d17be169425a7acf4e25e09e9de81a0f7300ed80c5ec656R86

proto/config/devconfig.proto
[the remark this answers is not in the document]
christoph-zededa: Thx. I added it here: https://github.com/lf-edge/eve-api/pull/70/files#diff-35ea4ed74707f6ad6d17be169425a7acf4e25e09e9de81a0f7300ed80c5ec656R86

proto/config/netcmn.proto
```diff
@@ -83,12 +83,25 @@ message ipspec {
...
+  // ntp and more_ntp are used to specify several NTP servers
   string ntp = 7;
+  repeated string more_ntp = 10;
   repeated string dns = 8;
 
   // for IPAM management when dhcp is turned on.
   // If none provided, system will default pool.
   ipRange dhcpRange = 9;
+
+  DhcpOptionsIgnore dhcp_options_ignore = 11;
```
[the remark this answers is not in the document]
christoph-zededa: I am not sure if we should put meaning in it if it is `nil` (because I don't know if other programming languages support this). Therefore if it is `nil` it should be handled as if the default value is set, i.e. `false`.
Therefore if `dhcp_options_ignore` is not defined, "use NTP from both DHCP and the static config".

If you want to use NTP from DHCP only, set `ntp` and `more_ntp` to empty strings.

Do we do this somewhere else? If yes, then it is perhaps better to do it the way you suggest.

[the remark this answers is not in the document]
milan-zededa: I see. No, I'm not aware of any place where we would check if struct (proto message) is set or unset.
Your approach makes sense to me now. I would maybe just suggest to add a comment (next to `dhcp_options_ignore`  ?) that even if `DhcpType` is `Client`, EVE still accepts static IP config and will merge it with the DHCP-provided one.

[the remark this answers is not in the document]
eriknordmark: In the protobuf wire encoding the fields with default values are not sent. Thus e.g., a boolean which is false, an integer of value zero, is not sent. As a result the receiver can not tell whether the sender explicitly set it to false or zero in those examples. I imagine this applies to a struct/embedded message as well; if all fields are zero it might not be sent at all.
I don't know if this is useful.

proto/config/netcmn.proto
```diff
@@ -83,12 +83,28 @@ message ipspec {
   string subnet = 3;
   string gateway = 5;
   string domain = 6;
+  // ntp and more_ntp are used to specify several NTP servers
```
eriknordmark, line 86: Presumably with ntp being the first and more_ntp being the rest aka the concatenation of ntp+more_ntp specifies all the NTP servers.

proto/config/netcmn.proto
```diff
@@ -83,12 +83,28 @@ message ipspec {
...
   string ntp = 7;
+  repeated string more_ntp = 10;
   repeated string dns = 8;
 
   // for IPAM management when dhcp is turned on.
   // If none provided, system will default pool.
   ipRange dhcpRange = 9;
+
+  // Even if DhcpType is Client, EVE still accepts static
+  // IP config and will merge it with the DHCP-provided one by default
```
eriknordmark, line 96: Note that the ZEDEDA controller doesn't appear to allow specifying some static info if the type is Client. I don't know what flexibility we need here (static IP, static gateway, static NTP servers, static DNS servers, static routes) when the type is client. 
Do we want to be able to specify the set of DHCP option codes to ignore? If so, can we use the DHCP option numbers instead of adding (one initially) booleans in DhcpOptionsIgnore?

proto/config/netcmn.proto
christoph-zededa: Hmm, I would not call it "all NTP servers" because there are possibly also NTP servers configured via DHCP.

eriknordmark: Correct. But it would be helpful to have a comment that it uses the union of ntp and more_ntp as the set of statically/manually configured NTP servers. And it wouldn't hurt to point out that this is due to originally only having a field for a single server.

proto/config/netcmn.proto
eriknordmark: Ah - I forgot about that. Then it makes sense to have booleans. But maybe the comment for  each boolean can come with a comment saying "this means the content of DHCP option X and DHCPv6 option Y are ignored" or something to that effect.

PR #71, opened 2024-10-17
vm : add vmResolution field to VmConfig

shjala:
The vm_resolution field is used to set the custom resolution for the VM running in FML mode. This is to move FmlCustomResolution from cloud-init to AppInstanceConfig.


proto/config/vm.proto
```diff
@@ -80,4 +80,6 @@ message VmConfig {
   bool enable_vnc_shim_vm = 22;
   // Boot mode: Legacy or UEFI
   VmBootMode boot_mode = 23;
+  // Custom resolution applies to VM that is runing in FML mode only.
+  string vmResolution = 24;
 }
```
OhmSpectator, line 85: As far as I remember, the config that comes in `vm.proto` becomes a part of FixedResources. As far as I remember, it means we cannot change this part of config after the App has started without purging the app. I think the resolution should be more dynamic. In this case it should become a part of `appconfig.proto`

PR #72, opened 2024-11-05
KubeClusterMetrics: node and app cpu/mem usage.

andrewd-zededa:
For kubevirt-eve images this adds per node and eve-user-app cpu/memory usage metrics.

Fix: add cluster_id to ZInfoKubeClusterUpdateStatus


proto/metrics/metrics.proto
```diff
@@ -710,3 +712,46 @@ message FlowlogCounters {
...
+
+  // Memory usage in bytes
+  uint32 memory_usage = 4;
+
+  // Memory usage in percentage
+  float memory_usage_percentage = 5;
+}
+
+// Resources used as reported by kubernetes api for pods running eve user apps.
+message KubeEVEAppPodMetrics {
```
zedi-pramodh, line 735: Why only pod metrics ? What about the VMI ? Do we still use pod metrics for VMI too ?

andrewd-zededa: Each vmi app has an associated pod (virt-launcher-*), vmi and vmirs objects in the eve-kube-app namespace.  

The vmi lists cpu cores, memory Gi resources,  and storage bytes requested.

The vmirs list cpu cores and memory Gi resources. 

The virt-launcher pod lists two containers virt-launcher-monitor (the vm) and virt-tail (logs).  Neither of these specifically list cores but instead a Millicore value which doesn't seem to match the core count.  My virt-launcher reports 200millicores which is 1/5th of a core instead of the vmi and vmirs which report 2 cores.  Using memory requests from the virt-launcher pod does seem to be the most complete view of all memory used by all containers associated with a running domain.

eriknordmark (approved):
LGTM

proto/metrics/metrics.proto
```diff
@@ -710,3 +712,49 @@ message FlowlogCounters {
...
+
+  // Memory usage in bytes
+  uint32 memory_usage = 4;
+
+  // Memory usage in percentage
+  float memory_usage_percentage = 5;
+}
+
+// Resources used as reported by kubernetes api for running eve user apps.
+message KubeEVEAppMetrics {
```
deitch, line 735: Why `KubeEVEAppMetrics` and not just `KubeAppMetrics`? This is running on EVE, and an EVE app in this context is a Kubernetes app. Do we need the extra `EVE` in it?

deitch (approved):
Not holding it up for the naming question, so approved, but if you agree on it, would be good to change.

proto/metrics/metrics.proto
andrewd-zededa: Yes I guess its a bit redundant, I'll shorten it

andrewd-zededa: Shorter `KubeAppMetrics` pushed up.

deitch: Thanks.

PR #76, opened 2024-11-27
attest : remove "storage" and "volume" from the field names

shjala:
Remove "storage" and "volume" from the field names, this is to make the field names more generic and not tied to storage or volume. This will make adding new keys to the attestation process easier. There has been no change in the wire format.


proto/attest/attest.proto
```diff
@@ -91,76 +91,76 @@ enum AttestVersionType {
...
 message ZAttestQuoteResp {
-  ZAttestResponseCode response = 1;  // Result of quote validation
-  bytes integrity_token = 2;         // Valid if response is ATTEST_RESPONSE_SUCCESS, contains integrity token for device to
-                                     // Cache and present in the next config request
-  repeated AttestVolumeKey keys = 3; // Valid if response is ATTEST_RESPONSE_SUCCESS, contains keys of type KeyType to unlock
-                                     // volume vault(s)
+  ZAttestResponseCode response = 1;    // Result of quote validation
+  bytes integrity_token = 2;           // Valid if response is ATTEST_RESPONSE_SUCCESS, contains integrity token for device to
+                                       // Cache and present in the next config request
+  repeated AttestKey storage_keys = 3; // Valid if response is ATTEST_RESPONSE_SUCCESS, contain keys (recived encrypted
```
eriknordmark, line 142: note typo "recived"

proto/attest/attest.proto
```diff
@@ -12,57 +12,57 @@ option java_package = "org.lfedge.eve.attest";
...
 // The message is assumed to be protected by signing envelope
 message ZAttestReq {
-  ZAttestReqType reqType = 1;  //type of the request
-  ZAttestQuote quote = 2;      //attestation quote msg
-  repeated org.lfedge.eve.certs.ZCert certs = 3;    //X509 certs in .PEM format, signed by device certificate
-  AttestStorageKeys storage_keys = 4;    //encrypted secrets to be saved by the Controller, like encrypted keys for the volume storage vaults
+  ZAttestReqType reqType = 1;                    // type of the request
+  ZAttestQuote quote = 2;                        // attestation quote msg
+  repeated org.lfedge.eve.certs.ZCert certs = 3; // X509 certs in .PEM format, signed by device certificate
+  AttestStorageKeys storage_keys = 4;            // encrypted secrets to be saved by the Controller, like encrypted keys for the volume storage vaults
```
eriknordmark, line 18: Shouldn't this field also have "Storage" removed?

proto/attest/attest.proto
```diff
@@ -12,57 +12,57 @@ option java_package = "org.lfedge.eve.attest";
...
 enum ZAttestReqType {
   ATTEST_REQ_NONE = 0;
-  ATTEST_REQ_CERT = 1;   //EVE X.509 certificates
-  ATTEST_REQ_NONCE = 2;  //nonce request to Controller
-  ATTEST_REQ_QUOTE = 3;  //quote msg
-  Z_ATTEST_REQ_TYPE_STORE_KEYS = 4; //to store device keys in Controller
+  ATTEST_REQ_CERT = 1;        // EVE X.509 certificates
+  ATTEST_REQ_NONCE = 2;       // nonce request to Controller
+  ATTEST_REQ_QUOTE = 3;       // quote msg from device
+  Z_ATTEST_REQ_TYPE_KEYS = 4; // to store device keys in Controller
```
eriknordmark, line 35: Here STORE is a verb having nothing to do with storage, thus I think you should keep it.

proto/attest/attest.proto
```diff
@@ -12,57 +12,57 @@ option java_package = "org.lfedge.eve.attest";
...
 enum ZAttestRespType {
   ATTEST_RESP_NONE = 0;
-  ATTEST_RESP_CERT = 1;       //response to cert msg
-  ATTEST_RESP_NONCE = 2;      //response to quote request
-  ATTEST_RESP_QUOTE_RESP = 3; //response to quote msg
-  Z_ATTEST_RESP_TYPE_STORE_KEYS = 4; //response to Z_ATTEST_REQ_TYPE_STORE_KEYS
+  ATTEST_RESP_CERT = 1;        // response to cert msg
+  ATTEST_RESP_NONCE = 2;       // response to quote request
+  ATTEST_RESP_QUOTE_RESP = 3;  // response to quote msg
+  Z_ATTEST_RESP_TYPE_KEYS = 4; //response to Z_ATTEST_REQ_TYPE_KEYS
```
eriknordmark, line 43: Ditto

proto/attest/attest.proto
shjala: corrected.

proto/attest/attest.proto
shjala: done.

proto/attest/attest.proto
shjala: done.

proto/attest/attest.proto
shjala: done.

eriknordmark (approved):
LGTM but we need to coordinate these changes with the controllers so that they get the changes and no surprises that they need to do the renames in their source code.

shjala:
> LGTM but we need to coordinate these changes with the controllers so that they get the changes and no surprises that they need to do the renames in their source code.

OK, lets have this pending until I've informed the relevant developers.

cshari-zededa:
The best time to do this change is when we bring in a new key to the attestation scheme.  There will be changes required on both the sides at that time, and we can absorb this change as well.

PR #78, opened 2025-01-02
add MAINTENANCE_MODE_REASON_TPM_ENCRYPTION_FAILURE

shjala:
We have seen errors like "point is not on the required curve" in the
past that prevents the device from upgrading because TPM fails to perform
an ECDH key agreement to generate the KEK and decrypt the vault key.
To catch these errors early, check if TPM is not functioning properly on
a critical path we set the device into maintenance mode with the reason
MAINTENANCE_MODE_REASON_TPM_ENCRYPTION_FAILURE.

This should go before https://github.com/lf-edge/eve/pull/4462


eriknordmark (approved):
LGTM

PR #79, opened 2025-01-03
Remove comment about maintenance mode reason

shjala:
Remove comment about maintenance mode reason, the API should be source of truth for the definition, not EVE, and as of https://github.com/lf-edge/eve/pull/4462 that is the case.


eriknordmark (commented):
Doesn't this belong to your other PR which uses the values in the API?

eriknordmark (approved):
Time to merge

PR #80, opened 2025-02-03
Maintenance mode changes

shjala:
1.Mark MaintenanceModeReason as deprecated.
2.Add tpm quote failure to maintenance mode reasons.

https://github.com/lf-edge/eve/pull/4462 needs this.


proto/info/info.proto
```diff
@@ -607,7 +607,7 @@ message ZInfoDevice {
 
   bool maintenance_mode = 38;
   // XXX to be deprecated in favor of maintenance_mode_reasons
-  MaintenanceModeReason maintenance_mode_reason = 39;
+  MaintenanceModeReason maintenance_mode_reason = 39 [deprecated=true];
```
eriknordmark, line 610: Does the added deprecated=true do anything to the ability to compile and use the old?
A controller would need to be able to compile and use both reason and reasons since old edge nodes will send the old one only.

shjala: No, it should juts show a warning when used in new code, from [Protobuf Documents](https://protobuf.dev/programming-guides/proto3/) :

> deprecated (field option): If set to true, indicates that the field is deprecated and should not be used by new code. In most languages this has no actual effect.

eriknordmark (approved):
LGTM (but someone needs to update the upload workflow since it fails).

eriknordmark:
Please rebase

shjala:
rebased.

PR #82, opened 2025-02-06
Add immediate snapshots

christoph-zededa:


OhmSpectator (approved):
Looks good to me, as far as @xyuria-zededa is fine with it.

PR #83, opened 2025-02-07
Patch Envelope Encryption Support

andrewd-zededa:
Add optional encryption to PatchEnvelope EveBinaryArtifact

To use encryption a controller should set EveBinaryArtifact.artifact_encryption_enabled
and supply CipherBlocks in the following fields

EveBinaryArtifact.metadata_cipher_data:
 Encryption for EveBinaryArtifact.artifactMetaData
 CipherBlock.cipherData should contain the encrypted serialized copy of an EncryptionBlock structure with
 EncryptionBlock.user = ENCRYPTION_BLOCK_USER_BINARY_ARTIFACT_METADATA
 and EncryptionBlock.encrypted_data set to the EveBinaryArtifact.artifactMetaData text.

If the PE EveBinaryArtifact contains an encrypted artifact
then it must include the encrypted artifact object
(InlineOpaqueBase64Data or ExternalOpaqueBinaryBlob) inside
an EncryptedInlineOpaqueBase64Data or EncryptedExternalOpaqueBinaryBlob
respectively.  
The EveBinaryArtifact.format field should remain set
to BASE64 or BINARYBLOB respectively, the EveBinaryArtifact.artifact_encryption_enabled
boolean directs that an Encrypted* type is used and must be decryted.
The CipherBlock.cipherData fields inside both Encrypted* binaryblob
types should contain the encrypted serialized copy of
InlineOpaqueBase64Data or ExternalOpaqueBinaryBlob respectively.


andrewd-zededa:
There is a yetus issue with the enum naming of PATCH_DECRYPTED.  I could fix it but then it wouldn't match the rest of the enum which uses PATCH_*

proto/info/patch_envelope.proto
```diff
@@ -33,6 +33,11 @@ enum EVE_PATCH_ENVELOPE_STATE {
     // Application instances are now allowed to fetch
     // contents
     PATCH_ACTIVE = 6;
+
+    // Artifact fields can be encrypted
+    // InlineOpaqueBase64Data: both base64Data and base64MetaData
+    // ExternalOpaqueBinaryBlob: only blobMetaData
+    PATCH_DECRYPTED = 7;
```
eriknordmark, line 40: Isn't this state implicit in PATCH_READY? If there was a decryption error after it was downloaded the state would move to PATCH_FAILED and otherwise to PATCH_READY?

proto/config/patch_envelope.proto
```diff
@@ -55,6 +56,12 @@ message EveBinaryArtifact {
     }
 
     optional string artifactMetaData = 4;
+
+    // Optional encryption for EveBinaryArtifacts
+    // Artifact fields which will be encrypted
+    // InlineOpaqueBase64Data: both base64Data and base64MetaData
+    // ExternalOpaqueBinaryBlob: only blobMetaData
```
eriknordmark, line 63: Why are we not encrypting all of InlineOpaqueBase64Data and all of ExternalOpaqueBinaryBlob? Should make it simpler to handle.

proto/config/patch_envelope.proto
```diff
@@ -55,6 +56,12 @@ message EveBinaryArtifact {
     }
 
     optional string artifactMetaData = 4;
+
```
eriknordmark, line 59: We have a SECRET type - will that be used?

There is a comment on lin 49  saying it is invalid, but if we want/need a type for this then maybe we need a separate SECRET_BASE64 and SECRET_BLOB?

udit-zededa:
@andrewd-zededa We will need cipher_data field for _InlineOpaqueBase64Data_: [base64Data](https://github.com/lf-edge/eve-api/blob/main/proto/config/patch_envelope.proto#L21) and [base64MetaData](https://github.com/lf-edge/eve-api/blob/main/proto/config/patch_envelope.proto#L28) and _ExternalOpaqueBinaryBlob_: [blobMetaData](https://github.com/lf-edge/eve-api/blob/main/proto/config/patch_envelope.proto#L41) as well.

CC: @zed-rishabh @gkodali-zededa

proto/info/patch_envelope.proto
andrewd-zededa: I can remove this enum state.

andrewd-zededa:
@uncleDecart The existence of the EveBinaryArtifact.cipher_data would serve as the indication that either binaryblob is encrypted.

proto/config/patch_envelope.proto
andrewd-zededa: Encryption of only a few sub-fields of the binaryblob structures was chosen to avoid defining a new parent structure for InlineOpaqueBase64Data and ExternalOpaqueBinaryBlob

andrewd-zededa:
After discussion this morning with @uncleDecart @udit-zededa @gkodali-zededa this PR has been modified to provide 4 separate CipherBlocks for each of the 4 fields being encrypted.  If the controller is encrypting all fields with the same CipherContext then all of the CipherBlocks included here (a multiplied across many artifacts in a Patch Envelope) should be able to reference the same CipherContext.  

Let me know of any concerns and I'll try to address them quickly.

proto/evecommon/acipherinfo.proto
```diff
@@ -62,6 +62,10 @@ message CipherBlock {
   bytes clearTextSha256 = 4;
 }
 
+enum EncryptionBlockUser {
+  EBU_UNSPECIFIED = 0;
```
eriknordmark, line 66: Since this is a new enum type please follow the yetis recommendations.

proto/evecommon/acipherinfo.proto
```diff
@@ -79,4 +83,11 @@ message EncryptionBlock {
   string cellular_net_password = 7;
   // Cluster token
   string cluster_token = 8;
+
+  // A set of fields to be used for future expansion of
+  // this object, if user is specified the encrypted data
+  // will be inside encrypted_data and to be used
+  // by the user specified.
+  EncryptionBlockUser user = 9;
```
eriknordmark, line 91: Do you have a use for these now? If not, why add them now?

eriknordmark (changes requested):
You also need to add an api capability enum value in info.proto so the controller can tell the device supports it

proto/config/patch_envelope.proto
eriknordmark: But how do we know the other current (and potentially future) fields in those structs do not need encryption?
It is a bit hard to explain e.g., that the inline PE is encrypted when only some fields of it are encrypted.

eriknordmark:
> After discussion this morning with @uncleDecart @udit-zededa @gkodali-zededa this PR has been modified to provide 4 separate CipherBlocks for each of the 4 fields being encrypted. If the controller is encrypting all fields with the same CipherContext then all of the CipherBlocks included here (a multiplied across many artifacts in a Patch Envelope) should be able to reference the same CipherContext.
> 
> Let me know of any concerns and I'll try to address them quickly.

See my concern above. I don't understand why defining a new struct is hard. Is the complexity in protobuf encoding a struct in the controller, then encrypting the encoded data? Or something else?

andrewd-zededa:
@eriknordmark What do you think about adding new structures to the "oneof binaryBlob" inside EveBinaryArtifact.  eg. EncryptedInlineOpaqueBase64Data and EncryptedExternalOpaqueBinaryBlob which contain a cipher block of the encrypted structure and then associated EVE_OPAQUE_OBJECT_CATEGORY enums ENCRYPTED_INLINE and ENCRYPTED_EXTERNAL?

andrewd-zededa:
@eriknordmark @uncleDecart I've pushed a new proposal here for patch envelope encryption and updated the commit and PR messages.  In this new proposal the EveBinaryArtifact message contains new oneof options to encrypt the entire InlineOpaqueBase64Data or ExternalOpaqueBinaryBlob and embed it in a cipher block.

I'd appreciate any reviews and comments.   @eriknordmark I'm trying to meet your concern above.

proto/config/patch_envelope.proto
```diff
@@ -8,12 +8,15 @@ option go_package = "github.com/lf-edge/eve-api/go/config";
 option java_package = "org.lfedge.eve.config";
 
 import "google/protobuf/timestamp.proto";
+import "evecommon/acipherinfo.proto";
 
 enum EVE_OPAQUE_OBJECT_CATEGORY {
     UNKOWN = 0;
     BASE64 = 1;
     BINARYBLOB = 2;
     SECRET = 3;
```
eriknordmark, line 17: What was the intent with the SECRET enum value @gkodali-zededa ? Is it currently used? If not, can we remove it now that we are addin the ENCRYPTED_*?

proto/evecommon/acipherinfo.proto
eriknordmark: How does it relate to the EncryptedInlineOpaqueBase64Data and EncryptedExternalOpaqueBinaryBlob messages?

proto/evecommon/acipherinfo.proto
```diff
@@ -79,4 +85,11 @@ message EncryptionBlock {
   string cellular_net_password = 7;
   // Cluster token
   string cluster_token = 8;
+
+  // A set of general fields to be used for expansion of
+  // this object, if user is specified the encrypted data
+  // will be inside encrypted_data and to be used
+  // by the user specified.
+  EncryptionBlockUser user = 9;
+  string encrypted_data = 10;
```
eriknordmark, line 94: This message has various clear text fields so it is odd to have a field named like this. 
And how will this be used?

eriknordmark:
> @eriknordmark What do you think about adding new structures to the "oneof binaryBlob" inside EveBinaryArtifact. eg. EncryptedInlineOpaqueBase64Data and EncryptedExternalOpaqueBinaryBlob which contain a cipher block of the encrypted structure and then associated EVE_OPAQUE_OBJECT_CATEGORY enums ENCRYPTED_INLINE and ENCRYPTED_EXTERNAL?

The new EncryptedInlineOpaqueBase64Data and EncryptedExternalOpaqueBinaryBlob messages makes sense to me, but I don't understand the other additions in this PR since I can't see how they'll be used. Will the "user" field be for serialized clear text of arbitrary messages? (such as EncryptedInlineOpaqueBase64Data and EncryptedExternalOpaqueBinaryBlob)

eriknordmark:
Note that the two commits are no longer cleanly separated into .proto file changes in one and generated files in the other. Makes sense to clean that up.

proto/evecommon/acipherinfo.proto
```diff
@@ -62,6 +62,12 @@ message CipherBlock {
   bytes clearTextSha256 = 4;
 }
 
+// User of the EncryptionBlock
+enum EncryptionBlockUser {
+  EBU_UNSPECIFIED = 0;
+  EBU_PATCH_ENVELOPE_BASE64 = 1;
+  EBU_PATCH_ENVELOPE_BINARYBLOB = 2;
+}
```
udit-zededa, line 70: @andrewd-zededa What I'm thinking for preparing encrypted `InlineOpaqueBase64Data` and `ExternalOpaqueBinaryBlob`, we will use same proto and marshal it and send it to EVE as cipherBlock, instead of using `EncryptionBlock` for it.
But yes for EveBinaryArtifact `artifactMetaData` we will require a new field inside EncryptionBlock.
@gkodali-zededa let me know your thoughts around it.

andrewd-zededa:
> Note that the two commits are no longer cleanly separated into .proto file changes in one and generated files in the other. Makes sense to clean that up.

Yeah I'll clean that up

proto/evecommon/acipherinfo.proto
andrewd-zededa: EncryptionBlock.encrypted_data is for users of CipherBlock.  In this case the EveBinaryArtifact.artifactMetaData.  The controller will:
-  allocate an EncryptionBlock with encrypted_data set to the artifactMetaData and user set to ENCRYPTION_BLOCK_USER_BINARY_ARTIFACT_METADATA
- the EncryptionBlock is then marshaled, encrypted and set in the into the cipherData of the CipherBlock.

andrewd-zededa:
@udit-zededa @gkodali-zededa 

EncryptionBlock.encrypted_data is for the EveBinaryArtifact.artifactMetaData.  Set the EncryptionBlock.encrypted_data to the artifactMetaData field, and then continue with marshal, encryption of the EncryptionBlock structure, and set into the CipherBlock.

proto/evecommon/acipherinfo.proto
andrewd-zededa: Use the new EncryptionBlock.encrypted_data and set user to ENCRYPTION_BLOCK_USER_BINARY_ARTIFACT_METADATA.

Or maybe I need to rename it to protected_data?

proto/config/patch_envelope.proto
gkodali-zededa: @eriknordmark Original intent was to add encrypted data support then, but then we decided to not do it. I do not think it is used today. This can be removed.

proto/evecommon/acipherinfo.proto
gkodali-zededa: @andrewd-zededa Controller can fill the `InlineOpaqueBase64Data`/`ExternalOpaqueBinaryBlob` and then use these structures to directly encode and encrypt. This encrypted data is directly placed inside `EncryptedInlineOpaqueBase64Data`/`EncryptedExternalOpaqueBinaryBlob` structures. With that, we won't need to place clear text data inside `EncryptionBlock` for this case.

proto/config/patch_envelope.proto
uncleDecart: Why can't we just use SECRET for both BASE64 and BINARYBLOB? Why do we need two different objects in the first place?

proto/evecommon/acipherinfo.proto
andrewd-zededa: @gkodali-zededa would the EncryptionBlock still used for the EveBinaryArtifact.artifactMetaData ?

proto/config/patch_envelope.proto
andrewd-zededa: @uncleDecart we can try setting SECRET for either case, eve will need to attempt to Unmarshal to both inline and external cases when it sees SECRET.

andrewd-zededa: @gkodali-zededa @uncleDecart my above comment is referencing this path in eve which breaks open the EveBinaryArtifact: https://github.com/lf-edge/eve/blob/5aff6b504b55c4bb24066c645743783fc11ddd3d/pkg/pillar/cmd/zedagent/parsepatchenvelopes.go#L81

proto/evecommon/acipherinfo.proto
andrewd-zededa: Comments added to EveBinaryArtifact.metadata_cipher_data to direct usage of EncryptionBlock fields.

proto/config/patch_envelope.proto
andrewd-zededa: Switched the latest push to using SECRET for both cases, documented in commit message and EVE_OPAQUE_OBJECT_CATEGORY enum.

andrewd-zededa:
PR description updated to match latest commit message.

andrewd-zededa:
@eriknordmark to address the earlier issue, api capability added: API_CAPABILITY_ENCRYPTED_PATCH_ENVELOPE

proto/config/patch_envelope.proto
andrewd-zededa: Latest commit has moved to use the SECRET type.

proto/config/patch_envelope.proto
gkodali-zededa: @andrewd-zededa How will EVE know if it should read the secret as EncryptedInlineOpaqueBase64Data or EncryptedExternalOpaqueBinaryBlob?

andrewd-zededa: @gkodali-zededa eve would need to check both EveBinaryArtifact.GetEncryptedInline() and EveBinaryArtifact.GetEncryptedVolumeref() for non nil.

gkodali-zededa: Thanks @andrewd-zededa!

gkodali-zededa (approved):
LGTM

proto/config/patch_envelope.proto
andrewd-zededa: One last change, hopefully.  After talking with @gkodali-zededa the usage of the SECRET format is no longer used.  Existing formats BASE64 and BINARYBLOB would be used for the encrypted types and the eve side will determine that the Encrypted* types (EncryptedInlineOpaqueBase64Data or EncryptedExternalOpaqueBinaryBlob) are used instead through the EveBinaryArtifact.artifact_encryption_enabled flag.

proto/config/patch_envelope.proto
```diff
@@ -45,16 +51,38 @@ message ExternalOpaqueBinaryBlob {
...
+}
+
 message EveBinaryArtifact {
     // base64 or binary blob only - secret type is invalid
     EVE_OPAQUE_OBJECT_CATEGORY format = 1;
 
     oneof binaryBlob {
         InlineOpaqueBase64Data inline = 2;
         ExternalOpaqueBinaryBlob volumeRef = 3;
+        EncryptedInlineOpaqueBase64Data encrypted_inline = 7;
```
eriknordmark, line 66: Why did you skip tag 4-6? If they were used before then please add them as deprecated.

andrewd-zededa: 4 needs to be skipped due to existing artifactMetaData field.  I only skipped 5 and 6 because I created metadata_cipher_data and artifact_encryption_enabled first.  I can re-id the new Encrypted* fields up to 5/6 and move the metadata_cipher_data to 7 and artifact_encryption_enabled to 8.

eriknordmark: Ah - hard to read but correct.

eriknordmark (approved):
LGTM but I don't know if you have already looked at the implementation complexity for controller(s) as well as EVE-OS; whether using separate types vs. a separate boolean field makes a difference.
If you are reasonably confident on both we can go ahead and merge this as is.

andrewd-zededa:
I believe @naiming-zededa will be handling the eve decryption side.  Do you have time to review this?

naiming-zededa:
LGTM @eriknordmark @andrewd-zededa

proto/config/patch_envelope.proto
uncleDecart: @andrewd-zededa could you please add comment as to why they're skipped? Because it's not obvious

proto/evecommon/acipherinfo.proto
uncleDecart: @andrewd-zededa how does this field correlate with artifact_encryption_enabled field? Why can't we use similar approach with oneof for metadata_cipher_data as well?

uncleDecart:
Before merging this I want to have a quick chat with @andrewd-zededa. I still have some questions about API and want to talk about implementation side of things

andrewd-zededa:
Talked with @uncleDecart this morning and made the following changes:
- comments to more effectively show the intentional skip in field IDs in EveBinaryArtifact.
- Removed EveBinaryArtifact.artifact_encryption_enabled as it is unnecessary.  The generated go code contains methods to determine existence of the encrypted types.

uncleDecart (approved):

udit-zededa (approved):

andrewd-zededa:
Pushed up a rebase to resolve conflicts in generated files.

andrewd-zededa:
Pushed up another rebase to resolve new conflicts in proto and generated files.

andrewd-zededa:
@eriknordmark are there any further changes needed to merge this?

PR #84, opened 2025-02-11
Add OEM Windows license key field to AppInstanceConfig

shjala:
Add OEM Windows license key field to AppInstanceConfig

If set to true the target application will receive the OEM Windows license key
embedded in the ACPI table (if any is present).


eriknordmark (approved):
LGTM

PR #85, opened 2025-02-12
Windows License Propagation Changes

shjala:
This PR brings the following changes :
1.  Revert "Add OEM Windows license key field to AppInstanceConfig" (not used yet so it is OK)
2. Update APICapability to indicate supports for Windows license passthrough.
3. Add enable_oem_win_license_key field to VmConfig (instead of AppInstanceConfig) to indicate that VM
  should receive Windows license embedded in ACPI tables .


proto/config/vm.proto
```diff
@@ -92,4 +92,6 @@ message VmConfig {
   // is limited. Depending on the application's operating system, achieving the desired network
   // interface order might not be possible.
   bool enforce_network_interface_order = 24;
+  // The app should receive the embedded Windows license key
+  bool oem_win_license_key = 25;
```
eriknordmark, line 96: Maybe it fits better here, but this change is unrelated to the info addition.

proto/info/info.proto
```diff
@@ -111,6 +111,7 @@ message ZInfoManufacturer {
   string biosVendor = 7;
   string biosVersion = 8;
   string biosReleaseDate = 9;
+  bool win_oem_licence_present = 10;
```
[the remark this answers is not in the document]
shjala: > With this boolean did you intend to also indicate that there is a license present in the ACPI tables?

yes, my intention was to set it if theres is licence in ACPI table. Should I remove it?

proto/config/vm.proto
shjala: EVE defines `"fixed resources", which means that the virtual machine must be restarted before changes to the field will take effect.`, So I thought it is better to be here.

proto/info/info.proto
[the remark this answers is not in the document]
eriknordmark: It would make sense to check how the ZEDEDA controller handles the API capabilties field, because it has a way to feed that to the UI so the UI can expose the features available for a given device. 
I don't know if having a separate field (even though it contains more information - that the license is actually present in the DMI tables) it might be more painful to propagate that information to the UI.

proto/config/vm.proto
eriknordmark: You can keep it here, but add some text to the commit and PR description about it.

eriknordmark:
Yetus failure about a field being deleted can be ignored, since that field was just added and has not yet been put in use  by any API user.

proto/info/info.proto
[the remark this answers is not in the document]
shjala: dropped it for now.

proto/config/vm.proto
shjala: added.

OhmSpectator (approved):
Looks pretty simple and fine

PR #86, opened 2025-02-13
Add an 'AppDeploymentType' to the AppInstanceConfig

naiming-zededa:
- for EVE to handle the docker compose logs and stats from the runtime,
      which needs to be handled differently such as IoT-Edge containers.


proto/config/appconfig.proto
```diff
@@ -235,3 +238,11 @@ message VolumeRef {
   // if mount_dir is empty then it will be mounted on /mnt
   string mount_dir = 3;
 }
+
+// AppRuntimeType - to indicate the types of Application Deployment
+// which may need special handling on the controller or on the device
```
eriknordmark, line 243: Presumably the controller can do special handling without it being in the EVE API, so drop "controller or" from the comment?

Can we document what IOT_EDGE implies (connect to port X for metrics and logs ...) and likewise for DOCKER?

Also, today the device handles Azure IoT Edge without any setting. How will we phase in setting 1 in the controllers and then having EVE-OS look for that value?

eriknordmark (commented):
Please split out the proto file in a separate config to make it easier to resolve the current (and potential future) conflicts.

proto/config/appconfig.proto
naiming-zededa: Ok, reword the comments
I removed the IOT_EDGE item, so this is only for docker compose and future extension if needed for special handling.
We'll document this on the eve doc side when submitting EVE side of the changes related to this API

eriknordmark:
> > Please split out the proto file in a separate config to make it easier to resolve the current (and potential future) conflicts.
> 
> yes, created another file appruntime.proto.

I meant to say "separate commit" but I clearly wasn't reading what I was typing ...
You still have a set of conflicts to resolve on the generated files.

naiming-zededa:
messed up this PR, create a new one #88

PR #88, opened 2025-02-19
Add an 'AppRuntimeType' to the AppInstanceConfig

naiming-zededa:
- for EVE to handle the docker compose logs and stats from the runtime,
  which needs to be handled differently such as IoT-Edge containers.
- messed up PR #86, redo this PR here.


eriknordmark (approved):
LGTM

PR #89, opened 2025-02-20
Ecc errors new info type

jsfakian:
- Added new types of ZInfoMsg protobuf messages for reporting the status of ECC memory


proto/info/info.proto
```diff
@@ -1366,3 +1369,20 @@ message ZInfoKubeCluster {
...
+  int64 ce_count = 2;
+  int64 ue_count = 3;
+}
+
+message ECCMemoryControllerInfo {
+  string controller_name = 1;
+  int64 ce_count = 2;
+  int64 ue_count = 3;
+  repeated DimmRankInfo ranks = 4;
+}
```
eriknordmark, line 1384: Please document what these fields are. rank, ce, ue, etc.

Is the repeated DimRankInfo one entry per dimm? Or for something else?

jsfakian: For the DimRankInfo we have one entry per dimm.

eriknordmark (changes requested):
We have a current message ZMetricMsg
I suggest you define a new message ZSlowMetricMsg and put these metrics in there.
(And make it be sent every 24 hours by default. with its own ConfigItem timer to adjust the frequency.)

proto/healthchecks/health_checks.proto
```diff
@@ -0,0 +1,50 @@
...
+// system hardware components. Each message typically contains status updates
+// for key hardware elements, such as ECC memory and storage devices. These
+// reports are generated at a low frequency (e.g., once every 24 hours) and
+// are not retransmitted in case of network failure, as their data remains
+// relatively stable.
+//
+// The message is secured within a TLS session, which is bound to the device
+// certificate to ensure integrity and authenticity.
+message ZHealthCheckMsg {
+  string devID = 1;
```
eriknordmark, line 24: For these new messages/fields can you use the suggested lower_snake_case (even though we are not changing the use in existing messages)

proto/healthchecks/health_checks.proto
```diff
@@ -0,0 +1,50 @@
...
+
+  google.protobuf.Timestamp atTimeStamp = 2;
```
eriknordmark, line 26: Ditto.

proto/healthchecks/health_checks.proto
```diff
@@ -0,0 +1,50 @@
...
+//
+// ZHealthCheckMsg encapsulates periodic health check reports for critical
+// system hardware components. Each message typically contains status updates
+// for key hardware elements, such as ECC memory and storage devices. These
+// reports are generated at a low frequency (e.g., once every 24 hours) and
+// are not retransmitted in case of network failure, as their data remains
+// relatively stable.
+//
+// The message is secured within a TLS session, which is bound to the device
+// certificate to ensure integrity and authenticity.
```
eriknordmark, line 22: Shouldn't we be using AuthContainer here where the AuthContainer is signed using the device identity?
(We don't seem to be using that for the Info and Metrics messages but it would make sense to introduce it for new messages like this one.)

proto/healthchecks/health_checks.proto
```diff
@@ -0,0 +1,50 @@
+// Copyright(c) 2025 Zededa, Inc.
+// All rights reserved.
+
+syntax = "proto3";
+
+import "google/protobuf/timestamp.proto";
+
+package org.lfedge.eve.healthchecks;
```
eriknordmark, line 8: FWIW you can ignore the buflint complaints for this line.

eriknordmark (commented):
Some nits plus the key design question whether we should use AuthContainer to sign these.

proto/healthchecks/health_checks.proto
jsfakian: It is a good idea to use the AuthContainer. However, I need some help implementing it. Can you give me some pointers? I saw how we do it for `CompoundEdgeDevConfig` in proto/config/compound_devconfig.proto, but I am unsure how to do it in our case.

eriknordmark: Sure I can provide pointers. I don't know if we have sending code in EVE-OS or only have the receiving code.
For now it makes sense to update the comments in this API PR.

eriknordmark (approved):
LGTM

PR #93, opened 2025-04-04
Extend CellularAccessPoint with LTE attach configuration

milan-zededa:
LTE connection consists of two IP bearers, the initial EPS bearer
and the default EPS bearer. Device must first establish the initial
bearer (LTE attach procedure, which ModemManager shows as a transition
from the `enabled/searching` to `registered` modem state) and then it
connects to a default bearer (transition from `registered` to `connected`).
Both bearers require PDP (Packet Data Protocol) context settings (APN,
ip-type, potentially username/password, etc.). Settings for the initial
bearer are typically provided by the modem profiles or the network, while
settings for the default bearer are user-configured. However, there are
cases, especially with private cellular networks or when changing SIM card
and moving to another operator for which the modem was not pre-configured,
where the modem may not provide the correct APN settings for the attach
procedure and it will fail to register into the network.
For these cases, it is useful to have the LTE attach parameters
user-configurable.

Additionally, we will now allow the IP-type to be configured for both
bearers. Currently, EVE sequentially attempts all IP types until
a connection is established, causing unnecessary delays. Specifying
the correct IP type in the configuration will streamline the process.


milan-zededa:
Moving to draft - will be adding some more fields to `ZCellularStatus` to report:
- modem profiles
- attach bearer settings (to check when not configured by the user)

proto/config/netconfig.proto
```diff
@@ -181,6 +187,16 @@ message CellularAccessPoint {
   // Not listed technologies will not be tried.
   // If empty, then modem will select RAT automatically.
   repeated org.lfedge.eve.common.RadioAccessTechnology preferred_rats = 7;
+  // The IP addressing type to use for the default bearer.
+  org.lfedge.eve.common.CellularIPType ip_type = 8;
```
eriknordmark, line 191: Is there a wildcard value for this which will cause the existing behavior (which is to try all until one works)?

proto/config/netconfig.proto
```diff
@@ -181,6 +187,16 @@ message CellularAccessPoint {
...
+  // Access Point Network for the attach (aka initial) bearer.
+  string attach_apn = 9;
+  // The IP addressing type to use for the attach bearer.
+  org.lfedge.eve.common.CellularIPType attach_ip_type = 10;
+  // Authentication protocol used for the attach bearer.
+  CellularAuthProtocol attach_auth_protocol = 11;
+  // Encrypted user credentials for the attach bearer (if required).
+  org.lfedge.eve.common.CipherBlock attach_cipher_data = 12;
```
eriknordmark, line 199: Is the assumption that the default and attach values have to all be specified?
Or is there some semantics that if only the default ones are specified, those will also be used for the attach?
Please document what happens if not both attach and default ones are specifiied.

milan-zededa: Attach is optional - added clarification under the comment for `string attach_apn = 9;`

proto/config/netconfig.proto
milan-zededa: The zero value that EVE will use when upgraded from an older version is `CELLULAR_IP_TYPE_UNSPECIFIED`.
Newer EVE will also use it when user does not specify the IP type.
I added comment for this value explaining that EVE will try every possible IP type in that case.

milan-zededa: I found out that the ip-type does not have to be specified in the bearer config, in which case the modem will use the pre-installed profiles to determine the right IP type for the network. I think this is better behavior going forward. I have seen many problems with our current approach of trying to connect with each IP-type with some timeout. It is problematic mostly because we do not know what is the ideal timeout. If it is too much, we are significantly delaying connectivity when trying wrong ip-type, and if it is too low, we may fail even with correct IP-type.
So now that it is configurable, I would keep the field as unspecified when not configured by the user and let the modem determine the proper value.
I have updated the comment for `CELLULAR_IP_TYPE_UNSPECIFIED`.

milan-zededa:
> I will also define API capability for this once #94 is merged (the "Single-Stack IP Network" capability must come first).

Done

eriknordmark (approved):
LGTM

proto/config/netconfig.proto
```diff
@@ -181,6 +187,22 @@ message CellularAccessPoint {
...
+  string attach_apn = 9;
+  // The IP addressing type to use for the attach bearer.
+  // Leave undefined if attach_apn is not specified.
+  org.lfedge.eve.common.CellularIPType attach_ip_type = 10;
+  // Authentication protocol used for the attach bearer.
+  // Leave undefined if attach_apn is not specified.
+  CellularAuthProtocol attach_auth_protocol = 11;
+  // Encrypted user credentials for the attach bearer (if required).
+  // Leave empty if attach_apn is not specified.
+  org.lfedge.eve.common.CipherBlock attach_cipher_data = 12;
```
milan-zededa, line 205: @eriknordmark Now I'm thinking if it was a good approach to add another CipherBlock field for attach username/password, or whether I should have used the already defined `org.lfedge.eve.common.CipherBlock cipher_data = 4` and add `cellular_net_attach_username` and `cellular_net_attach_password` next to [existing cellular fields](https://github.com/lf-edge/eve-api/blob/main/proto/evecommon/acipherinfo.proto#L81-L84C10) in the EncryptionBlock.
Which approach is preferred?
We can still change this, it is not yet used.

eriknordmark: I guess a single cipherblock in the object avoids some complexity since in theory each cipherblock can refer to a different controller encryption certificate, thus one can fail and the other succeed when there is some missing cert.

PR #92, opened 2025-04-03
Initial proto objects for metric server hosted in a runtime app instance

andrewd-zededa:
A webserver may be deployed inside an eve app instance which hosts
http endpoints returning nested application instance metrics.
These objects are returned in the body of the http GET so that
eve knows which application instance ids are running nested
inside the runtime and the metrics for each of those respective
nested app instances.

Migrated from lf-edge/eve-tools which was originally PR: https://github.com/lf-edge/eve-tools/pull/11
Following up on discussion in lf-edge/eve PR: https://github.com/lf-edge/eve/pull/4716

PR #95, opened 2025-04-23
Restore eve-tools field case to nestedappinstancemetric

andrewd-zededa:
NestedAppMetrics fields are protojson marshaled and unmarshaled.
Case changes cause compatibility breakages.

See PR #92


andrewd-zededa:
@naiming-zededa PR created for revert.

naiming-zededa (commented):
LGTM. This avoids the compatibility issue between eve versions and agents.

eriknordmark (approved):
LGTM
I didn't realize this was used with protojson, which means that changing names means a change to the wire format (unlike with protobuf wire encoding).

PR #96, opened 2025-05-02
Merge attach and default bearer user credentials into a single CipherBlock

milan-zededa:
To reduce configuration size and avoid the complexity of managing
two CipherBlock instances, credentials for both attach and default
bearers are now included in a single CipherBlock.

This is a breaking change, but it's acceptable because the attach
bearer configuration fields were added recently and are not yet in use.

Discussed here: https://github.com/lf-edge/eve-api/pull/93#discussion_r2069111492


eriknordmark (approved):
LGTM

PR #98, opened 2025-05-07
Allow sending a compressed yaml registration manifest to the cluster for download.

andrewd-zededa:
This should be used to cluster management, not app instance deployment.  Compression must be gzip.


andrewd-zededa:
@xyuria-zededa Please review this change.

eriknordmark (commented):
Is there a design document with examples of how this will be used, and any compatibility issues with the cluster token usage?

xyuria-zededa:
@andrewd-zededa This new field will be empty for the standard edge-node clusters and non empty for the zks edge-node clusters, correct?

proto/config/edge_node_cluster.proto
```diff
@@ -30,4 +30,8 @@ message EdgeNodeCluster {
 
   // encrypted token string, use edge-node TPM to decrypt
   org.lfedge.eve.common.CipherBlock encrypted_cluster_token = 7;
+
+  // URL to download a yaml manifest for a cluster
+  // These manifests should not define app-instance deployments.
+  org.lfedge.eve.common.CipherBlock encrypted_manifest_yaml_url = 8;
```
milan-zededa, line 36: It might make sense to avoid two cipher blocks and use one for both the token and the yaml url.
See: https://github.com/lf-edge/eve-api/pull/93#discussion_r2069111492

andrewd-zededa:
> @andrewd-zededa This new field will be empty for the standard edge-node clusters and non empty for the zks edge-node clusters, correct?

Correct

proto/config/edge_node_cluster.proto
andrewd-zededa: After further review the encryption of this field is not needed and will be removed in an upcoming push to this pr.

andrewd-zededa:
> Is there a design document with examples of how this will be used, and any compatibility issues with the cluster token usage?

@eriknordmark this new url field "registration_manifest_yaml_url" is optional and separate from the existing cluster token.

andrewd-zededa:
@zedi-pramodh @xyuria-zededa Following our conversation I've changed this to the new model.  The controller must have the registration manifest compressed via gzip and provide it here.

andrewd-zededa:
@eriknordmark The goal of this PR still aims to provide the yaml manifest in config from the controller.  The earlier version of this PR required the controller providing a URL to EVE and the responsibility would be on EVE to download it.  The new plan as just pushed up requires the controller get the yaml manifest, compress down to ~1KB and embed this in config.

proto/config/edge_node_cluster.proto
```diff
@@ -30,4 +30,8 @@ message EdgeNodeCluster {
 
   // encrypted token string, use edge-node TPM to decrypt
   org.lfedge.eve.common.CipherBlock encrypted_cluster_token = 7;
+
+  // GZIP Compressed yaml manifest used for completing registration
+  // into a controller.
+  bytes gzip_registration_manifest_yaml = 8;
```
eriknordmark, line 36: Does this need to be encrypted?

eriknordmark (commented):
In addition to the encrypted/cleartest question, will the controller always include both the encrypted_cluster_token and the new field, or will it need to know whether a particular instance of EVE-OS supports the new field?

You need to add an enum value to the api_capabilities if it is the latter.

proto/config/edge_node_cluster.proto
zedi-pramodh: It is just a URL string not the contents so I do not see reason to encrypt, unless the current plan is to send the content too,  @andrewd-zededa ?

zedi-pramodh: I am confused why we need to get the content from cloud ? Why can't we just get the URL string ? Is the plan to download using eve API ? But we download many kubernetes objects outside eve api.

xyuria-zededa:
> In addition to the encrypted/cleartest question, will the controller always include both the encrypted_cluster_token and the new field, or will it need to know whether a particular instance of EVE-OS supports the new field?
> 
> You need to add an enum value to the api_capabilities if it is the latter.

We already check this on the controller side. We check that optcapabilities field include "kubevirt" option.  The additional field will be used only for the zks clusters and will be empty for the standard edge-node clusters. 
@zedi-pramodh Andrew wanted to pass a content of the manifest instead of passing a manifest url due to some security concerns.

zedi-pramodh:
I am not 100% sure of the security concern here. It's just a URL string and it comes secured with our TLS encryption.  Assuming someone got hold of the URL and first of all he/she needs to know that should be executed on kubernetes cluster.  They cannot even do DDoS because that can be run only once.

Extremely narrow window I see is the following:

1) User created cluster on controller
2) EVE pulled the config (URL)
3) Hacker somehow watching network exactly at sametime, saw the URL
4) EVE suddenly lost connection to controller with in millisecs of config download.
5) Meanwhile hacker executed the URL on his/her kubernetes cluster.

How about encrypting URL with device cert instead of passing the content ?

xyuria-zededa:
> I am not 100% sure of the security concern here. It's just a URL string and it comes secured with our TLS encryption. Assuming someone got hold of the URL and first of all he/she needs to know that should be executed on kubernetes cluster. They cannot even do DDoS because that can be run only once.
> 
> Extremely narrow window I see is the following:
> 
>     1. User created cluster on controller
> 
>     2. EVE pulled the config (URL)
> 
>     3. Hacker somehow watching network exactly at sametime, saw the URL
> 
>     4. EVE suddenly lost connection to controller with in millisecs of config download.
> 
>     5. Meanwhile hacker executed the URL on his/her kubernetes cluster.
> 
> 
> How about encrypting URL with device cert instead of passing the content ?

I am totally fine with passing the manifest in an encrypted string. That's what i suggested from the beginning.

zedi-pramodh:
@andrewd-zededa I am suggesting to send URL to EVE and download there not in controller. Why should controller download manifest for 1000s of individual clusters ? It should be responsibility of eve, IMO

xyuria-zededa:
> > > I am not 100% sure of the security concern here. It's just a URL string and it comes secured with our TLS encryption. Assuming someone got hold of the URL and first of all he/she needs to know that should be executed on kubernetes cluster. They cannot even do DDoS because that can be run only once.
> > > Extremely narrow window I see is the following:
> > > ```
> > > 1. User created cluster on controller
> > > 
> > > 2. EVE pulled the config (URL)
> > > 
> > > 3. Hacker somehow watching network exactly at sametime, saw the URL
> > > 
> > > 4. EVE suddenly lost connection to controller with in millisecs of config download.
> > > 
> > > 5. Meanwhile hacker executed the URL on his/her kubernetes cluster.
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > How about encrypting URL with device cert instead of passing the content ?
> > 
> > 
> > I am totally fine with passing the manifest in an encrypted string. That's what i suggested from the beginning.
> 
> @xyuria-zededa That is the proposed plan here, controller will get the yaml string, compress with gzip from ~5KB down to output ~1KB and input that compressed binary data here.

Sorry for the confusion. By "I am totally fine with passing the manifest in an encrypted string. " i meant passing the manifest url not the manifest content.

zedi-pramodh:
It's from docker hub AFAIK.

eriknordmark:
> It's from docker hub AFAIK.

So that's an argument that the URL approach will be problematic, and we're better off including the manifest itself.
But it isn't clear to me from the thread above that you all are on the same page about the approach.

zedi-pramodh:
Yes we are not on same page and I have call with @andrewd-zededa on Monday to discuss this approach. But basically the URL is to Controller to download manifest yaml. The manifest yaml contains the images which are in docker hub. So I am not seeing any reason for controller to create a manifest content for 1000s of clusters and send to every cluster. But we can discuss.

proto/config/edge_node_cluster.proto
andrewd-zededa: This PR follows the plan as discussed on Saturday Morning 5/24 to switch from passing the url -> embedding manifest body (~1KB compressed).  We can discuss today before I revert back to the old PR.

zedi-pramodh:
We discussed this internally and decided to keep this PR as is and get the manifest content from controller in compressed format. That will eliminate any corner case security issues and the content is only seen by eve devices.

eriknordmark:
> The URL proposal as mentioned by @zedi-pramodh I already had implemented in the old version of this PR and we discussed on 5/24 about switching to this style of passing the manifest body.  I can switch back but I thought we were in agreement on moving to passing the manifest body in eve-api.

Ok sounds like we're on the same page.

> - The manifest (yaml) contains a secret for the cluster.  The entire eve-api config is protected in flight to eve with the TLS wrapper so encrypting this field only adds a second layer.  I think any security concerns are more focused on some other client guessing the endpoint path and pulling the secret themselves.

Note that all the secrets and potential secrets in the config (such as cloud unit, pes, wifi creds) are protected using object encryption in addition to TLS. Based on what you are saying I think we should do the same for this new field.

andrewd-zededa:
@eriknordmark I'll prep a change to encrypt this field.

eriknordmark:
> @eriknordmark @xyuria-zededa.  I moved this field "gzip_registration_manifest_yaml" into the EncryptionBlock.  Can we share that config.EdgeNodeCluster.CipherBlock for both the cluster_token as well as gzip_registration_manifest_yaml?  Or do we require two separate CipherBlocks?

What @milan-zededa has done and suggested elsewhere is a single CipherBlock in a message and multiple fields in that cipher block filled in.

eriknordmark (commented):
@milan-zededa should we comment where we use a CipherBlock which fields can be filled in?

Other than such potential comments this looks good to me.

milan-zededa:
> @eriknordmark @xyuria-zededa. I moved this field "gzip_registration_manifest_yaml" into the EncryptionBlock. Can we share that config.EdgeNodeCluster.CipherBlock for both the cluster_token as well as gzip_registration_manifest_yaml? Or do we require two separate CipherBlocks?

We did use the approach of sharing single CipherBlock for multiple cellular attributes and there were no issues on the cloud or the EVE side.

eriknordmark (approved):
LGTM

zedi-pramodh:
Can this be merged if no other issues, we need this code to be available for controller side work.

PR #99, opened 2025-05-08
devconfig: upload datastore for collect-info@LOC

christoph-zededa:
this enables to specify a datastore where eve will upload the collect-info tarball to if requested to do by the local operator console

python and go changes will follow


proto/profile/local_profile.proto
```diff
@@ -138,6 +138,8 @@ message AppCommand {
       // A subsequent action to start the app will start it with a pristine runtime state.
       // This command will purge ALL volumes used by the application.
       COMMAND_PURGE = 2;
+      // Application instance will be shut down.
+      COMMAND_SHUTDOWN = 3;
```
eriknordmark, line 142: What is the intended use? 
Two concerns:
1. This sounds like a change to the intended state (that the application should be HALTED forever) and we limit the commands to things which are more like operations (such as a reboot or a purge). What would happen if the controller sends updated configuration with activate=true before/after/at the same time as this API is used to send a shutdown command?
2.  It is odd to have a shutdown and not a corresponding start command.

christoph-zededa: Actually I see it might not be needed, because the user can just use the local profile server instead: https://github.com/lf-edge/eve-api/blob/main/PROFILE.md#appinfo

proto/config/devconfig.proto
```diff
@@ -27,6 +27,7 @@ import "auth/auth.proto";
 message LOCConfig {
   // If set indicates URL of the Local Operator Console (LOC)
   string loc_url = 1;
+  string datastore_collect_info_id = 2; // UUID of the datastore to upload the collect-info tarball to - only DsHttp and DsHttps is currently possible
```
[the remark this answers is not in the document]
eriknordmark: So all of the devices which use the same datastore will put there files in the same directory?
That seems suboptimal, unless we add the device UUID to the tar filename generated by collect-info.

proto/profile/local_profile.proto
eriknordmark: > 1. Isn't the use of LOC to only use it if the controller is not available? That's how I understand https://github.com/lf-edge/eve/blob/master/pkg/pillar/cmd/zedagent/handleconfig.go#L873

No, they are used in parallel.
The LOC and LPS can be viewed as synonyms for related concepts. If you want to distinguish them, the LPS API is defined in this local-profile.proto and it included non-declarative commands for a local operator.
The LOC might refer to the fact that EVE can fetch the full EdgeDevConfig (signed by the controller) from two sources - either the controller or a local server - and it picks the most recent version. That does not allow any local modifications by design. 
In addition in this general space we have the ability to send metrics and info (aka deployment status) to a local server in addition to the controller (and we should probably add the ability to send logs to a local server as well.)

Note that the LPS API does not have the ability to reboot a device - only restart (and purge etc) app instances.
It is only your proposed app instance shutdown command which I find problematic.

proto/profile/local_profile.proto
```diff
@@ -195,6 +197,15 @@ message LocalDevCmd {
...
       // the edge node or someone manually power cycling the edge node.)
       COMMAND_SHUTDOWN_POWEROFF = 2;
+      // REBOOT: Edge node will initiate a graceful reboot.
+      // The state of the application instances (HALTING, HALTED, etc) can be
+      // observed since is sent to the api/v1/appinfo API endpoint as normal.
+      // Any local profile server is shut down after all the other app instances
+      // have halted.
+      COMMAND_REBOOT = 3;
+      // COLLECT_INFO: Edge node will call the collect-info.sh script with
+      // the datastore parameters (see datastore_collect_info_id)
```
europaul, line 207: I think you should mention that EVE will also send the results to the pre-configured datastore

proto/profile/local_profile.proto
christoph-zededa: To me it looks like this is the feature availability in LOC and LPS:

Feature | LOC | LPS | 
--- | --- | --- 
collect-info | O | O
device reboot |  O | O
application shutdown | O | X

So, do we need "application shutdown" in LOC? Because whoever wants this can just use LPS.

proto/profile/local_profile.proto
christoph-zededa: like this?

europaul: oh, so you put the upload logic into `collect-info.sh`? then all good :+1:

proto/profile/local_profile.proto
eriknordmark: Don't they use the same protobuf for the commands?

proto/config/devconfig.proto
```diff
@@ -27,6 +27,10 @@ import "auth/auth.proto";
 message LOCConfig {
   // If set indicates URL of the Local Operator Console (LOC)
   string loc_url = 1;
+  // UUID of the datastore to upload the collect-info tarball to - only DsHttp and DsHttps is currently possible
+  // the directory is specified by dpath in DatastoreConfig and the filename is the standard collect-info.sh
+  // filename, f.e.: eve-info-v34-2025-05-13-08-53-52.tar.gz
+  string datastore_collect_info_id = 2;
```
zed-rishabh, line 33: Can we include a list of datastores here? This would be helpful if the user wants to configure multiple datastores as part of a fallback mechanism—similar to how it's handled in the app image download flow. If one datastore is unavailable, the system can automatically try the next one in the list.

proto/profile/local_profile.proto
christoph-zededa: Yes, but according to https://help.zededa.com/hc/en-us/articles/7168577376795-Manage-App-Instances-with-the-Offline-Profile-Server#h_01JMD6FYZ72GKN5JBF5BMWQ3K7 - there is "App Shutdown".

proto/config/devconfig.proto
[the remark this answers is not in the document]
christoph-zededa: I updated the comment that it contains the UUID of the device.

proto/profile/local_profile.proto
```diff
@@ -138,6 +138,10 @@ message AppCommand {
       // A subsequent action to start the app will start it with a pristine runtime state.
       // This command will purge ALL volumes used by the application.
       COMMAND_PURGE = 2;
+      // Application instance will be shut down.
+      COMMAND_STOP = 3;
+      // Application instance will be started.
+      COMMAND_START = 4;
```
eriknordmark, line 144: Seems like there is no request for these commands (and their semantics would be odd when the device has intermittent connectivity to the controller.)

So I think we should drop these from the PR.

eriknordmark (changes requested):
Please remove the START and STOP app instance commands.

christoph-zededa:
> Please remove the START and STOP app instance commands.

Done.

I also renamed `COMMAND_REBOOT` to `COMMAND_SHUTDOWN_REBOOT`, because it is more in line with `COMMAND_SHUTDOWN_POWEROFF`

europaul:
> I also renamed COMMAND_REBOOT to COMMAND_SHUTDOWN_REBOOT, because it is more in line with COMMAND_SHUTDOWN_POWEROFF

@christoph-zededa I don't think it's a good idea. After all you are just rebooting the device, so it's kinda implied that the applications will be rebooted as well. SHUTDOWN_REBOOT sounds to me like the apps will be shut down and might not come back up after the device reboot.

europaul:
> > > I also renamed COMMAND_REBOOT to COMMAND_SHUTDOWN_REBOOT, because it is more in line with COMMAND_SHUTDOWN_POWEROFF
> > 
> > @christoph-zededa I don't think it's a good idea. After all you are just rebooting the device, so it's kinda implied that the applications will be rebooted as well. SHUTDOWN_REBOOT sounds to me like the apps will be shut down and might not come back up after the device reboot.
> 
> @europaul What will be the name of the command to force reboot without first shutting down the apps?

@christoph-zededa is there really a difference between shutting down the apps first and then rebooting or just rebooting?

christoph-zededa:
> @christoph-zededa is there really a difference between shutting down the apps first and then rebooting or just rebooting?

`COMMAND_SHUTDOWN_REBOOT` to me is `shutdown -r now`, while `COMMAND_REBOOT` sounds more like `reboot -f` (https://linux.die.net/man/8/reboot )

The difference is that the apps get a signal for poweroff and only then the machine gets rebooted versus the machine gets immediately rebooted.

europaul:
> > @christoph-zededa is there really a difference between shutting down the apps first and then rebooting or just rebooting?
> 
> `COMMAND_SHUTDOWN_REBOOT` to me is `shutdown -r now`, while `COMMAND_REBOOT` sounds more like `reboot -f` (https://linux.die.net/man/8/reboot )
> 
> The difference is that the apps get a signal for poweroff and only then the machine gets rebooted versus the machine gets immediately rebooted.

maybe `COMMAND_GRACEFUL_REBOOT` would be a good name? sounds less confusing to me

christoph-zededa:
> maybe `COMMAND_GRACEFUL_REBOOT` would be a good name? sounds less confusing to me

Okay, let's see who and what will complain when I rename `COMMAND_SHUTDOWN_POWEROFF` to `COMMAND_GRACEFUL_POWEROFF`

europaul:
> > > maybe `COMMAND_GRACEFUL_REBOOT` would be a good name? sounds less confusing to me
> > 
> > Okay, let's see who and what will complain when I rename `COMMAND_SHUTDOWN_POWEROFF` to `COMMAND_GRACEFUL_POWEROFF`
> 
> @europaul Yetus does not like to rename it, so one would be named `SHUTDOWN` the other one `GRACEFUL`. Or should we ignore yetus?

SHUTDOWN_POWEROFF sounds okay to me, I just find SHUTDOWN_REBOOT very confusing

europaul (approved):

proto/profile/local_profile.proto
```diff
@@ -194,7 +194,17 @@ message LocalDevCmd {
       // been shut down the edge node will also power off. Note that the power
       // on will need to be done through other local means (UPS power cycling
       // the edge node or someone manually power cycling the edge node.)
-      COMMAND_SHUTDOWN_POWEROFF = 2;
+      COMMAND_GRACEFUL_POWEROFF = 2;
```
eriknordmark, line 197: If we do this rename we need to notify all of our users that there is a change - no binary compatibility issue  thus no need for them to change, but if they later rebuild their applications which use the LPS API they have to do the rename in their source code.

I'm not sure it is worth-while figuring out how to write such a notice and explain it in a way where it doesn't cause confusion.

Also, we'd need to update the sample client code for the LPS and LOC API, but that is less of a concern for me. User confusion is the biggest concern.

christoph-zededa: I added a comment to explain it a bit better: https://github.com/lf-edge/eve-api/pull/99/files#diff-07ca59289357fd001106a6107bab9b5b3ee69b07d3a2693974649985d29953a2R197

IMO whoever updates it, sees the compile error and then hopefully finds this comment and is able to make the change.

I think it is better to have the user confusion now only once than having a constant confusion why we mix `GRACEFUL` and `SHUTDOWN`.

christoph-zededa: It also might be a good test to see who screams at us for doing this ;-)

christoph-zededa: I added `COMMAND_SHUTDOWN_POWEROFF` (and setting the option `option allow_alias = true;`) and it is now marked as deprecated.

christoph-zededa:
> Do you know why buflint complains? buflint:Enum option "allow_alias" on enum "Command" must be false.

https://buf.build/docs/breaking/rules/#enum_value_same_name - it is not really clear to me, because the new `LocalDevCmd` is still compatible with the old `LocalDevCmd`.

eriknordmark:
> > Do you know why buflint complains? buflint:Enum option "allow_alias" on enum "Command" must be false.
> 
> https://buf.build/docs/breaking/rules/#enum_value_same_name - it is not really clear to me, because the new `LocalDevCmd` is still compatible with the old `LocalDevCmd`.

OK, let's ignore it.
Please check if the LPS or LOC API clients generate any errors or warnings if you recompile them against the new version of the API.

PR #100, opened 2025-05-29
Add Device Info Capability of Edgeview Authentication

naiming-zededa:
- add Capability of Edgeview-Authentication, we will use that in Edgeview-UI. If policy requires authentication but the EVE version does not support that, we'll not allow Edgeview on the device.
- details see https://github.com/lf-edge/eve/blob/master/docs/EDGEVIEW-CONTAINER-API.md
- add python and go generated files


eriknordmark (approved):
LGTM. 
Would it make sense to add some reference to some edge view markdown file in the EVE repo in the comment?

naiming-zededa:
> LGTM. Would it make sense to add some reference to some edge view markdown file in the EVE repo in the comment?

updated the commit message to reference to the md file.

eriknordmark (commented):
Info is fine. But please add the info to the comment in the proto file.

naiming-zededa:
> Info is fine. But please add the info to the comment in the proto file.

@eriknordmark updated the proto comments.

PR #101, opened 2025-06-10
Deprecate event_log in attest.proto

shjala:
Deprecate the `event_log` field in `attest.proto`, we should not parse the TPM event log to omit some information or compress it, it should be send "as is" to the controller to be replayed and verified by the controller.


APIv2.md
```diff
@@ -276,7 +276,12 @@ If `reqType` is `ATTEST_REQ_CERT`, `certs` should be filled with certificates us
 
 If `reqType` is `ATTEST_REQ_NONCE`, no other field needs to be set in `ZAttestReq`.
 
-If `reqType` is `ATTEST_REQ_QUOTE`, then `quote` MUST be populated with the following information: `attestData` containing [TPMS_ATTEST structure defined in TCG](https://trustedcomputinggroup.org/wp-content/uploads/TPM-Rev-2.0-Part-2-Structures-01.38.pdf), `signature` containing signature of `TPMS_ATTEST` by the attestation signing key, `pcr_values` containing values of PCR 0-15, `event_log` containing TPM Event Log (aka measurement logs) with each event entry of the format `TpmEventLogEntry`,  `versions` containing current versions of various software components running on the device(e.g. EVE, UEFI etc), `gps_info` containing GPS coordinates of device's geo-location.
+If `reqType` is `ATTEST_REQ_QUOTE`, then `quote` MUST be populated with the following
+information: `attestData` containing [TPMS_ATTEST structure defined in TCG](https://trustedcomputinggroup.org/wp-content/uploads/TPM-Rev-2.0-Part-2-Structures-01.38.pdf), `signature` containing signature of
```
eriknordmark, line 280: You have a markdownlint line length issue here.

PR #102, opened 2025-06-12
Add option to enable LLDP forwarding inside a network instance

milan-zededa:
This commit adds a new configuration option, `forward_lldp`,
to enable forwarding of LLDP (Link Layer Discovery Protocol)
frames within a network instance.

By default, the Linux bridge implementation conforms to IEEE 802.1D
and applies strict filtering of certain reserved Ethernet multicast
destination addresses — specifically those in the range
`01:80:C2:00:00:00` to `01:80:C2:00:00:0F`. LLDP frames, which use EtherType
`0x88cc` and destination address `01:80:C2:00:00:0E`, fall within this range
and are therefore not forwarded by the bridge.

While this filtering is appropriate in many physical switching scenarios
to avoid propagating link-local control protocols, it can be problematic in virtualized
environments. When the Linux bridge is used to interconnect virtual machines
or containers, the ability to forward LLDP frames between endpoints is often
required.

The `forward_lldp` flag provides explicit control over this behavior,
allowing LLDP frames to be forwarded through the network instance when needed.
By default, LLDP forwarding will continue to be disabled for backward
compatibility.


milan-zededa:
Alternatively, we can just have a config-property defined for this (applied to all deployed network instances) and avoid EVE API change. WDYT @eriknordmark, @uncleDecart, @naiming-zededa ?

eriknordmark (approved):
LGTM

PR #103, opened 2025-06-23
Add field for disabling vTPM in VM config

shjala:
Introduce a new field `vtpm_enabled` in the VM configuration protocol. This field allows users to enable or disable the virtual TPM (vTPM) feature for a VM. By default, the vTPM is enabled.

_This is necessary to ensure that VMs can continue to function if there are issues with vTPM when the VM is not using vTPM._


shjala:
> LGTM Do we document the VmConfig fields anywhere outside of the protobuf definition?

Not that I know of, in EVE we document global configs but not per VM config AFAIK.

proto/config/vm.proto
```diff
@@ -94,4 +94,7 @@ message VmConfig {
   bool enforce_network_interface_order = 24;
   // The app should receive the Windows license embedded in ACPI tables
   bool enable_oem_win_license_key = 25;
+  // Disable vTPM for the VM, this defaults to false according to the
+  // proto3 specification.
+  bool disable_vtpm = 26;
```
OhmSpectator, line 99: I think we are trying to keep such bool fields with "enable_" semantics. The "disable_" one makes the code a little bit confusing.

proto/config/vm.proto
```diff
@@ -94,4 +94,7 @@ message VmConfig {
   bool enforce_network_interface_order = 24;
   // The app should receive the Windows license embedded in ACPI tables
   bool enable_oem_win_license_key = 25;
+  // Disable vTPM for the VM, this defaults to false according to the
```
OhmSpectator, line 97: Are you planning to make it configurable for an application at runtime? I mean, if you put it here in `vm.proto`, it will become part of the `FixedResources` in `AppInstanceConfig`, which means the user would need to update the app template and perform a Purge & Update to change this setting. That might be the right approach for your use case.

But if you want it to be more dynamic - allowing changes without a purge - then it should go to https://github.com/lf-edge/eve-api/blob/f2951f18534ec5c4b218104d5279575dae03de0a/proto/config/appconfig.proto#L106

OhmSpectator: As @eriknordmark has already approved, I guess you are doing it right. Just want to be sure)

OhmSpectator: Makes sense!

proto/config/vm.proto
OhmSpectator: Ok, got it! I like the approach @milan-zededa suggested, but I'm also fine with the current approach.

PR #106, opened 2025-07-28
Nested App Runtime Filesystem Usage Metrics

andrewd-zededa:
A runtime deployed onto EVE-OS possibly as a VM app instance
will contain a persistent filesystem used to store container
app images and volumes for instances of those apps.  Publish
the storage metric from that runtime with each nested/dependent
app's storage usage in this shared filesystem.


andrewd-zededa:
I think the remaining yetus errors are not valid

README_nestedapp.md
```diff
@@ -43,3 +43,16 @@ Return codes:
 Response:
 
 The body will contain the NestedAppMetrics structure in json form.
+
+### Storage Metric
+
+GET /api/v1/storagemetrics/runtime
```
eriknordmark, line 49: I don't understand the design choice to make this a single endpoint vs. the nestedf-app-id above given that each NestedAppRuntimeDiskMetric contains the app uuid.

In addition, in the info message this is repeated but here it is only included once. 
So I'm a bit confused.

andrewd-zededa: The repeated NestedAppRuntimeDiskMetric object is to handle multiple runtimes on a single EVE-OS node (any app instance where the AppInstanceConfig.runtime_type!=APP_RUNTIME_TYPE_UNSPECIFIED), I can amend the readme and comment there.  

Each NestedAppRuntimeDiskMetric is associated with a runtime persistent filesystem and provides the usage snapshot at a runtime scope which contains 1 or more nested app ids sharing the fs.  EVE would only issue a single get to the /api/v1/storagemetrics/runtime endpoint for each runtime app instance.

eriknordmark (approved):
So for a single compose VM running 3 compose files with a total of 7 containers would there a single nested storage metrics returned to eve? Or one per volume used by each container?

PR #110, opened 2025-08-20
network adapter info: replace LocalModification with generic PortConfigSource

milan-zededa:
This PR makes small enhancement to changes introduced by https://github.com/lf-edge/eve-api/pull/108
Given that these new API fields are not being used yet by any EVE version, we can ignore any backward-incompatibility warnings.

Description of changes:
- Removed `LocalModification` message, previously used just to distinguish
  LPS-modified port configs from others.
- Added `PortConfigSource` message to describe the origin of a port configuration.
- Introduced `ConfigOrigin` enum to enumerate all possible config origins:
  controller, bootstrap, override, last-resort, TUI, LOC and LPS.
- `PortConfigSource` includes a `submitted_at` timestamp reflecting when the
  configuration for this port was originally created or submitted at its source.
  For LPS-modified configs, this timestamp represents when the port-level
  modification was received. This timestamp is per-port and may differ between
  ports within the same underlying network configuration. Unlike
  `DevicePortStatus.timePriority`, `submitted_at` reflects the actual creation/submission
  time, not a synthetic priority value.


proto/info/info.proto
```diff
@@ -828,22 +828,53 @@ message DevicePort {
...
+  // Manually created JSON config injected into the installer.
+  NETWORK_CONFIG_ORIGIN_OVERRIDE = 3;
+  // Fallback configuration automatically generated to enable DHCP on all Ethernet ports.
+  // Used when no network config is available or when "network.fallback.any.eth" is enabled
+  // and none of the existing configs provide controller connectivity.
+  NETWORK_CONFIG_ORIGIN_LASTRESORT = 4;
+  // Config entered via the terminal UI (TUI).
+  NETWORK_CONFIG_ORIGIN_TUI = 5;
+  // Config changes made locally through the Local Profile Server (LPS).
+  NETWORK_CONFIG_ORIGIN_LPS = 6;
```
christoph-zededa, line 852: does it make to differentiate between LPS and LOC?

f.e. we do here: https://github.com/lf-edge/eve/blob/master/pkg/pillar/cmd/zedagent/zedagent.go#L287

proto/info/info.proto
```diff
@@ -828,22 +828,53 @@ message DevicePort {
   WirelessStatus wireless_status = 33; // defined for cellular/WiFi ports only
   uint32 mtu = 34; // MTU of the port
 
-  // Local modification status for this port.
-  // If modified, the configuration shown reflects local changes and
-  // represents the actual state of the port, not just a copy of the
-  // controller-submitted configuration.
-  LocalModification local_modification = 40;
```
christoph-zededa, line 835: What is the advantage of inlining this message?

The disadvantage is that it makes it harder to read it.

christoph-zededa (approved):

proto/info/info.proto
milan-zededa: Could you please clarify what you mean here (considering that you commented on the removed code)?

proto/info/info.proto
milan-zededa: Makes sense, added LOC as another possible origin.

proto/info/info.proto
christoph-zededa: all fine; seems I misread - I thought you have put the fields of `LocalModification` directly into `DevicePort`.

proto/info/info.proto
```diff
@@ -836,22 +836,56 @@ message DevicePort {
...
+message PortConfigSource {
+  // Indicates where EVE obtained the network config.
+  NetworkConfigOrigin origin = 1;
+
+  // Timestamp when EVE received or generated this configuration.
+  //
+  // Notes:
+  // - DevicePortStatus.timePriority contains the reception timestamp for controller-sourced
+  //   configs. For other sources, this value may not reflect the actual time the configuration
+  //   was generated or received.
```
eriknordmark, line 877: "received" seems a bit underspecified since unchanged configuration is "received" from the controller every 60 seconds by default, and there can be a change to the configuration from the controller at time X but that has no change to this port.

So is it received_change_at or received_at with the comment saying that it is the most recent change to this port?

PR #112, opened 2025-08-25
Move netcmn.proto to evecommon to break import cycle

milan-zededa:
In this PR, `netcmn.proto` has been moved under the `evecommon` package (where it
arguably should have been from the start) to resolve an import cycle
between the `config` and `profile` packages.

The cycle appeared because `CompoundConfig` (for LOC), defined in
`config`, imports LPS command messages from `profile`. After [a recent
PR](https://github.com/lf-edge/eve-api/pull/108) added a payload definition for a new `network` LPS endpoint, the
`profile` package also began importing `config` (for shared network
types), creating a cycle.

This change does not break wire compatibility: protobuf package names
are not encoded on the wire, so the serialized representation of the
moved messages remains unchanged. Only import paths for generated code
are affected, so users of `eve-api` will need to update their imports.
I tested EVE with these changes against controller built with older
`eve-api` to confirm this.

An alternative would have been to move `compound_devconfig.proto` into
a separate package (e.g. `proto/loc`), but common network types naturally
belong in `evecommon` anyway.


eriknordmark (approved):
LGTM (and I agree the renames has no wire encoding implications).

But needs to be rebased on master.

milan-zededa:
> LGTM (and I agree the renames has no wire encoding implications).
> 
> But needs to be rebased on master.

Done

PR #114, opened 2025-08-27
Adjust LPS network endpoint payload definitions

milan-zededa:
While implementing the new LPS network endpoint, I realized the payload definition could be improved. 
I made a few small adjustments, which I hope will be the final ones...

This work is independent of https://github.com/lf-edge/eve/pull/5196.  
That EVE PR can be merged to unblock https://github.com/lf-edge/eve/pull/5186.  
These adjustments will be propagated to EVE in a future PR, together with the implementation.

- Moved `PortConfigSource` into `evecommon/netcmn.proto` so it can be used
  by both `info/info.proto` and `profile/network.proto`.
- Replaced the unnecessary `NetworkConfig` wrapper in `profile/network.proto`
  with a direct `repeated NetworkPortConfig` field.
- Removed/renamed all references to "local config *change*" -- instead, the local
  configuration was defined to be fully declarative in nature
- Introduced `LocalNetworkConfigInfo` to report the state of the local config,
  including any errors:
    * global `error_message` indicates problems with the config as a whole,
      such as invalid token, malformed payloads or missing logical labels.
    * Per-port errors are still reported via `NetworkPortConfig.error_message`.
- Extended `NetworkConfigTestingStatus` with:
   * `testing_in_progress` boolean to indicate if controller connectivity
       testing is still ongoing.
   * `testing_phase` string providing optional context about why testing
       is in progress (e.g., waiting for interface, DHCP IP, or DNS).
- Renamed `logical_name` to `logical_label` in `profile/network.proto`
  for consistency with naming used in other proto files.
- Renamed `NetworkPortAddresses.usb_adapter` to `NetworkPortAddresses.usb_address`
  (the previous field name was a mistake)
- Added missing probing configuration into CellularConfig (in `profile/network.proto`)
- Updated documentation and comments in `profile/network.proto` for clarity.


milan-zededa:
I'm still making small adjustments as I'm working on the implementation, so I will keep this as a draft for now.

proto/profile/network.proto
```diff
@@ -66,53 +92,15 @@ message NetworkConfigChange {
...
-  // Error message describing why the last submitted configuration was
-  // not fully/partially applied.
-  // This may include validation errors for configurations received from either
-  // the controller or LPS, or permission-related errors for configuration changes
-  // requested by LPS. Empty if the last submitted configuration was applied successfully.
-  // For port-specific validation/permission errors, see `NetworkPortConfig.error_message`
-  string error_message = 4;
+  // Ports for which locally applied configuration should be removed,
+  // reverting to the controller-provided configuration.
+  // Identified by port logical labels.
```
eriknordmark, line 102: This doesn't sound like a very declarative way to express things. 
Raises question like for how long something should be included in reverted_ports. Forever if the LPS has ever made an override for that port? Until it sees something coming back for EVE?

An alternative approach would be to have LPS have to include all the ports (which it is allowed to change - presumable the LPS knows that from some message) in NetworkConfigChange and for the ones which it has no overrride it sets a boolean override = false.

proto/profile/network.proto
```diff
@@ -66,53 +92,15 @@ message NetworkConfigChange {
...
-  // Manually provided network-only config in JSON format and embedded into the EVE installer.
-  CONFIG_SOURCE_OVERRIDE = 3;
-  // Exported device config from the controller with the controller's signature,
-  // imported into the Local Operator Console (LOC), and fetched by the device over HTTPS.
-  CONFIG_SOURCE_LOC = 4;
-  // Created using a Local Profile Server (LPS) application and fetched
-  // by the device using the LPS API (currently HTTP-only).
-  CONFIG_SOURCE_LPS = 5;
-}
+  // Network configuration changes to be applied to one or more ports.
```
eriknordmark, line 95: "changes" and "NetworkConfigChange" sounds imperative and not declarative. Change since when?
See below for an alternative but if we go with a declarative approach then we shouldn't use "change" anywhere but instead use "local" and/or "override".

milan-zededa: OK, done. And yes, there is a per-port `error_message` defined inside `NetworkInfo`, which EVE publishes to LPS. This error message will be used to report permission, validation, etc. errors.

eriknordmark (approved):
LGTM

PR #115, opened 2025-09-01
Add message validation and some minor changes

shjala:
Amongst the other small changes (check commits), the only change that worth describing is the newly added annotations and message validation rules. Consider this start of an effort to move specification from textual description to unambiguous and validatable specification that lives in code. 

Using varios protobuf validation and annotation libraries we can precisely define EVE-APIs, including the content-type, endpoint path, accepted message type, returned type, HTTP status codes, schemes, custom headers, etc, etc. ~These definitions end up in swagger files and can be used to automatically generate tests for example~ , **_moved to a seperate PR_**. 
```
option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_swagger) = {
    schemes: HTTPS;
    consumes: ["application/x-proto-binary"]
    produces: ["application/x-proto-binary"]
};

service RegisterService {
    rpc RegisterDevice(ZRegisterMsg) returns (google.protobuf.Empty) {
        option (google.api.http) = {
            post: "/api/v1/edgedevice/register"
            body: "*"
        };
    option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_operation) = {
        summary: "Register a new Device for the first time."
        external_docs: {
            url: "https://github.com/lf-edge/eve-api/blob/main/APIv2.md"
            description: "API for Device to Controller communications"
        }
        operation_id: "RegisterDevice"
        responses: {
            key: "200"
            value: {
                description: "Repeat valid registration of Device"
            }
        }
[...]
```

In addition, we can also define the expected message, in form of validation rules for each field: 
```
string serial = 3 [
    (validate.rules).string = {
      min_len: 1,                   // Must not be empty
      max_len: 256,                 // Reasonable serial number length limit
      pattern: "^[a-zA-Z0-9_-]+$"   // Alphanumeric, underscore, hyphen only
    }
  ];
```

These rules results in automatic validation rule generation : 
```go
	if l := utf8.RuneCountInString(m.GetSerial()); l < 1 || l > 256 {
		err := ZRegisterMsgValidationError{
			field:  "Serial",
			reason: "value length must be between 1 and 256 runes, inclusive",
		}
		if !all {
			return err
		}
		errors = append(errors, err)
	}

	if !_ZRegisterMsg_Serial_Pattern.MatchString(m.GetSerial()) {
		err := ZRegisterMsgValidationError{
			field:  "Serial",
			reason: "value does not match regex pattern \"^[a-zA-Z0-9_-]+$\"",
		}
		if !all {
			return err
		}
		errors = append(errors, err)
	}
```

and can be used to validate received messages without guesswork and custom validation codes: 
```go
	regReq := &register.ZRegisterMsg{}
	if err := proto.Unmarshal(body, regReq); err != nil {
		return fmt.Errorf("Failed to unmarshal message: %v\n", err)
	}
	if err := regReq.Validate(); err != nil {
		return fmt.Errorf("Validation failed: %v\n", err)
	}
```

In this PR only register message is annotated and enriched with validations rules, more to come.

### DIFF
--- a/OBJECT-SIGNING.md
+++ b/OBJECT-SIGNING.md
@@ -34,7 +34,8 @@ The `AuthContainer` wrapper is constructed by:
 
 1. Put the above payload in the `protectedPayload` field
 1. Compute Sha256 over the above payload
-1. Compute the ECDSA signature of that sha, and place it in the `signatureHash` field
+1. Compute the ECDSA signature of that sha, and place it in the `signatureHash`
+field, the signature MUST be raw concatenation of r and s.
 1. To identify the sender, place a truncated sha of the sender's certificate in `senderCertHash`
 1. Place the algorithm used for the truncated sha in the `algo` field. Currently it's either a SHA256-32bytes or SHA256-16bytes. Note that `senderCertHash` is just for a lookup at the receiver, hence it can be truncated to be a lot shorter without any security implications.
 1. For the case when the receiver might not be able to identify the sender using just the hash, place the full sender's certificate in the `senderCert` field. This is the base64 standard encoding of the PEM format of the certificate. (This is used for the `register` API during onboarding.)


proto/register/register.proto
```diff
@@ -7,16 +7,114 @@ package org.lfedge.eve.register;
 option go_package  = "github.com/lf-edge/eve-api/go/register";
 option java_package = "org.lfedge.eve.register";
 
-// This is the request payload for POST /api/v1/edgeDevice/register
+// buf:lint:ignore
+import "validate/validate.proto";
```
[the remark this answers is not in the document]
deitch: @shjala where is it downloading it from, and what step is downloading it?

.gitignore
```diff
@@ -1 +1,3 @@
 .cache/
+go/go.sum
```
deitch, line 2: Why exclude this? Don't we want it in the git-published package, so that others can see and import a complete package?

deitch (commented):
I had a few questions, commented inline.

proto/register/register.proto
```diff
@@ -7,16 +7,114 @@ package org.lfedge.eve.register;
...
+
+// All POST requests MUST have the mime type set of application/x-proto-binary.
+// All responses with a body MUST have the mime type set of application/x-proto-binary.
+option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_swagger) = {
+    schemes: HTTPS;
+    consumes: ["application/x-proto-binary"]
+    produces: ["application/x-proto-binary"]
+};
+
+service RegisterService {
```
deitch, line 23: Is this the first time we are including the API endpoint service in the protobuf? Is this a pattern we intend to follow?

.gitignore
shjala: I don't know I though we build in the container so it doesn't matter, I will add it back.

proto/register/register.proto
deitch: > we can debate it

We probably should. Personally, I am in favour (probably strongly, too). I think that having the API paths and content (messages) defined in our official neutral API location (not inside a controller or client) is the right thing to do (so I agree with you). But it only helps if we adopt it, otherwise it gets out of sync.

My recommendation (take it for what it is worth, feel free to disagree, even "strongly" 😁 ) is to separate these out. Keep message validation in here, it is plenty for a PR. Then open a PR adding the paths, include interested folks who would adopt it: EVE, Adam (OSS controller), Eden (I think it has it somewhere), Zedcloud (commercial controller). Get agreement on "yes, this is good and we will adopt it" without holding up the good validation work here.

proto/register/register.proto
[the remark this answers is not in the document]
deitch: Oh that is interesting. it is in the `.devcontainer/Dockerfile`, not in the general build. Does it matter? Ca new always assume that the devcontainer will be used?

Oh, I see, it is referenced explicitly in the Makefile, so even if someone doesn't use devcontainers, it still will get referenced. That makes sense.

proto/register/register.proto
shjala: Makes sense, I'll move the API definition to another standalone PR.

shjala:
Update : 

Add commit 2a78cda288178f047ad7785079bc8a638a8f5a82, to explicitly mention how signature is formed,  apparently raw concatenation of R and S is called IEEE P.1363 format, but since the standard is not public and I couldn't find it I refrained from mentioning the standard.

eriknordmark (approved):
LGTM

shjala:
> A line length yetus issue to fix. I don't know if the buflint issues are expected or if they indicate a problem to fix

bulint issue is expected.

PR #143, opened 2026-04-15
Add bond adapter status and metrics

milan-zededa:
Publish runtime status and metrics for Linux bond adapters:

Info (`DevicePort.bond_status`):
- Bond mode currently applied
- Currently active bond-member logical label (for: active-backup, TLB, ALB modes)
- MII link monitoring status (enabled, polling interval, up/down delay)
- ARP link monitoring status (enabled, polling interval, IP targets, missed max)
- LACP status for 802.3ad mode (LACP rate, active aggregator ID,
  number of ports, partner MAC, actor/partner keys)
- Per-member status for all modes: logical label, MII up state,
  and LACP-specific info (aggregator ID, churn states) for 802.3ad

Metrics (`deviceMetric.bond_metrics`):
- Per-member link failure count
- Per-member LACP counters for 802.3ad mode (actor/partner churned counts)

Move `BondMode` and `LacpRate` enums from config (`devmodel.proto`) to
evecommon (`netcmn.proto`) so they can be shared between config and info
packages. This does not break backward compatibility of serialized data
since protobuf identifies enum values by their numeric value, not the
package path -- only import paths in consuming code need to be updated
(**ignore Yetus complaining about this change**).


proto/info/info.proto
```diff
@@ -798,6 +799,9 @@ message DevicePort {
 
   // Device-reported 802.1X (PNAC) runtime status for this port.
   PNACStatus pnac_status = 50;
+
+  // Runtime status of the bond, published only for bond adapters.
```
eriknordmark: Is "bond adapter" one which is part of/eslaved under a bond? Or the top of the bond? Please clarify comment.

proto/metrics/metrics.proto
```diff
@@ -796,3 +799,29 @@ message PNACMetrics {
   // Number of received EAPOL frames with incorrect length or truncated payload.
   uint64 eap_length_error_frames_rx = 10;
 }
+
+// BondMetrics contains metrics for a single bond adapter.
+message BondMetrics {
+  // Logical label identifying the bond adapter.
```
eriknordmark, line 805: Comment is clear here due to context.

proto/metrics/metrics.proto
```diff
@@ -796,3 +799,29 @@ message PNACMetrics {
...
+  string logicallabel = 1;
+  // Per-slave metrics.
+  repeated BondSlaveMetrics slaves = 2;
```
eriknordmark: I wonder whether it makes sense to do a global replace from "BondSlave" to "BondMember" or something similar.

proto/info/info.proto
milan-zededa: Actually both, updated comment to:
```
// Runtime status of the bond interface, including bond-level information as well as
// per-member details for adapters that are part of the bond.
```

proto/metrics/metrics.proto
milan-zededa: Done

milan-zededa: We also call them "members" in the controller, so it makes sense (only Linux still uses "slave").

eriknordmark (approved):
LGTM

PR #144, opened 2026-04-23
LPS network endpoint enhancements

milan-zededa:
# Summary

Three related additions to the Local Profile Server (LPS) API, centered on
the `/api/v1/network` endpoint and a new optional signaling channel,
aimed at **improving LPS-user experience**:

1. **New `GET /api/v1/signal` streaming endpoint.** A long-lived,
   optional endpoint that lets LPS notify EVE immediately when a
   configuration change becomes pending for any of the existing LPS
   endpoints, instead of waiting up to a minute (or whatever polling interval
   is configured) for the next periodic poll. This reduces the delay between
   a user submitting a configuration change via LPS and being notified that
   it has been applied, improving the overall user experience.
   See `proto/profile/signal.proto` and the `### Signal` section in `PROFILE.md`.

2. **`NetworkPortConfig.local_modifications_allowed` (field 20).**
   Mirrors the controller-provisioned
   `SystemAdapter.allow_local_modifications` flag. EVE reports this
   per-port in `NetworkInfo.latest_config[*]` so LPS can distinguish
   ports that accept local configuration from those managed exclusively
   by the controller -- without having to infer it by trial-and-error
   via `error_message`. Ignored by EVE when set by LPS on
   `LocalNetworkConfig.ports`.

3. **New `NetworkPortStatus` message, exposed as
   `NetworkInfo.port_status` (field 6).** Reports the current
   kernel-observed runtime state of each network port: link up/down,
   MAC address, assigned IPs (CIDR), active default routers, effective
   DNS servers and search domain, NTP servers in use, and applied MTU.
   Distinct from the existing `latest_config` / `fallback_config` /
   `local_config` fields, which describe what *should* be configured;
   this describes what is *currently* in effect.

## The Signal endpoint in more detail

- Method: `GET /api/v1/signal`.
- Return codes: `200` (streaming body follows), `404` (not implemented).
- Wire format: NDJSON (`application/x-ndjson`), one
  `profile.Signal` per line encoded with canonical protobuf JSON and
  terminated by `\n`.
- Implementation is **optional** -- LPS returning `404` means "no signal
  channel, keep polling"; EVE throttles its reconnect attempts (for this signal
  endpoint) in that case (~1/hour) so LPS can start supporting the endpoint
  later without a device reconfiguration.
- On each new connection, if any changes are already pending, LPS emits
  an initial `Signal` immediately covering them.
- The stream carries no `server_token`; tokens remain on the
  config-carrying response bodies of the referenced endpoints. A
  `Signal` never carries configuration itself, it only prompts EVE to
  poll the existing, token-protected endpoints, so the channel does
  not need its own authentication to remain safe.

Full normative spec is in `PROFILE.md`.


PROFILE.md
```diff
@@ -483,6 +483,10 @@ the `404` code.
 * Status of the local network configuration submitted previously
   by the Local Profile Server (LPS), indicating whether it was successfully
   applied or if errors occurred.
+* Per-port runtime status. This is the current kernel-observed state of each
+  network port (link up/down, MAC address, assigned IPs, effective DNS servers etc.)
```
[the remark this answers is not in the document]
milan-zededa: Done.

proto/profile/network.proto
```diff
@@ -43,6 +43,13 @@ message NetworkInfo {
   // Ports not listed here are using either the latest controller config
   // or the fallback config, as indicated by `NetworkPortConfig.config_applied`.
   LocalNetworkConfigInfo local_config = 5;
+
+  // Per-port runtime status observed on the device (kernel interface,
+  // routing table, resolver state). Unlike the NetworkPortConfig variants
+  // above, which describe what *should* be configure, this field
```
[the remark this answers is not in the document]
milan-zededa: Done

rene (approved):
LGTM

uncleDecart (approved):
LGTM, who has the priority on the config change? Do we keep timestamp somewhere? I.e. say we change at LPS and at the cloud, which changes would be used? How would it be communicated ?

milan-zededa:
> LGTM, who has the priority on the config change? Do we keep timestamp somewhere? I.e. say we change at LPS and at the cloud, which changes would be used? How would it be communicated ?

@uncleDecart 
See the “Behaviour” section under
[https://github.com/lf-edge/eve-api/blob/main/PROFILE.md#network](https://github.com/lf-edge/eve-api/blob/main/PROFILE.md#network)

If configuration changes made via LPS are permitted for a given port, the LPS-provided configuration takes highest priority. Any configuration later submitted through the controller is ignored (shadowed) as long as the LPS configuration remains in effect. This continues until either the LPS configuration is removed (by the LPS-user), or the controller-user revokes the permission for LPS to modify the port configuration.

milan-zededa:
@uncleDecart 

> How would it be communicated ?

See: https://github.com/lf-edge/eve-api/blob/main/proto/info/info.proto#L797-L798
And: https://github.com/lf-edge/eve-api/blob/main/proto/evecommon/netcmn.proto#L329-L356
And finally: https://github.com/lf-edge/eve-api/blob/main/proto/evecommon/netcmn.proto#L306-L327

uncleDecart:
> > LGTM, who has the priority on the config change? Do we keep timestamp somewhere? I.e. say we change at LPS and at the cloud, which changes would be used? How would it be communicated ?
> 
> @uncleDecart See the “Behaviour” section under https://github.com/lf-edge/eve-api/blob/main/PROFILE.md#network
> 
> If configuration changes made via LPS are permitted for a given port, the LPS-provided configuration takes highest priority. Any configuration later submitted through the controller is ignored (shadowed) as long as the LPS configuration remains in effect. This continues until either the LPS configuration is removed (by the LPS-user), or the controller-user revokes the permission for LPS to modify the port configuration.

So if LPS configuration fails we fallback to controller? In case if there was upgrade of EVE which changed fields and that led to failure of LPS configuration application we get to controller? Do we notify user when this happens?

milan-zededa:
> > > LGTM, who has the priority on the config change? Do we keep timestamp somewhere? I.e. say we change at LPS and at the cloud, which changes would be used? How would it be communicated ?
> > 
> > 
> > @uncleDecart See the “Behaviour” section under https://github.com/lf-edge/eve-api/blob/main/PROFILE.md#network
> > If configuration changes made via LPS are permitted for a given port, the LPS-provided configuration takes highest priority. Any configuration later submitted through the controller is ignored (shadowed) as long as the LPS configuration remains in effect. This continues until either the LPS configuration is removed (by the LPS-user), or the controller-user revokes the permission for LPS to modify the port configuration.
> 
> So if LPS configuration fails we fallback to controller? In case if there was upgrade of EVE which changed fields and that led to failure of LPS configuration application we get to controller? Do we notify user when this happens?

Yes, this follows the standard fallback procedure. If the highest-priority network configuration fails to provide working controller connectivity, the system attempts a lower-priority configuration, either from a different source or an older version from the same source, if available.
And we do expose the status of connectivity testing and any fallback configuration to LPS: https://github.com/lf-edge/eve-api/blob/main/proto/profile/network.proto#L28-L37
As a side note, an EVE upgrade should not disrupt network configuration. We maintain backward compatibility for DevicePortConfig (just like for any other persisted pubsub publication).

proto/profile/signal.proto
```diff
@@ -0,0 +1,52 @@
...
+// pending for the existing LPS endpoints. Upon receiving a Signal, EVE
+// immediately polls each endpoint listed in `pending_changes`, bypassing
+// the normal periodic polling cadence.
+//
+// The signal channel is strictly a latency optimization. Periodic polling
+// of the other LPS endpoints continues unchanged as the correctness
+// guarantee: if the signal stream is unavailable, broken, or not
+// implemented by LPS, EVE transparently falls back to polling only.
+//
+// Wire format: NDJSON (application/x-ndjson). Each Signal is encoded using
```
eriknordmark, line 22: Is this NDJSON signalling payload approach something standardized or commonly used?

PROFILE.md
```diff
@@ -542,6 +546,113 @@ Behavior:
...
+Whenever LPS has one or more configuration updates ready for EVE, it emits
+a [Signal](./proto/profile/signal.proto) message listing the affected
+endpoints. Upon receiving a signal, EVE immediately polls each listed
+endpoint, bypassing its normal periodic cadence.
+
+Implementing this endpoint is OPTIONAL. The signal channel is strictly
+a latency optimization: periodic polling of the other LPS endpoints continues
+unchanged as the correctness guarantee. If LPS does not implement the signal
+endpoint, or if the signal stream is temporarily unavailable or broken, EVE
+transparently falls back to polling only and behavior is otherwise unchanged.
```
eriknordmark, line 562: What about a case where the http connection fails after a while (due to a firewall or NAT or something else). Will EVE detect this and immediately reconnect? Are we running with TCP keepalive on this connection?

It might make sense to say something about that here.

PROFILE.md
```diff
@@ -542,6 +546,113 @@ Behavior:
...
+  responsibility (see below).
+* When LPS needs to terminate the stream (e.g., shutdown), it MUST close
+  the response cleanly.
+
+EVE behavior:
+
+* EVE MUST enable TCP keepalive on the signal connection with an interval
+  suitable for timely detection of a dead peer (approximately 60 seconds
+  is recommended). This is the sole mechanism for detecting a silently
+  broken connection; there is no application-layer heartbeat.
```
eriknordmark, line 628: The answer is here - but might make sense to state that in above context about "broken" LPS.

proto/profile/signal.proto
milan-zededa: When server-side signaling is required, it’s more common to use WebSockets or Server-Sent Events (SSE). However, since the LPS was already designed as a simple HTTP server, it made more sense to keep the approach lightweight and minimize changes to existing implementations by using newline-delimited JSON (NDJSON) streaming.

I did consider SSE, but it is generally more suited to browser-based clients. In our case, the client is a backend service, where parsing JSON is more natural, especially in Go. Additionally, since JSON is already used as the payload format across other LPS endpoints, NDJSON keeps the design consistent.

PROFILE.md
milan-zededa: Added (brief) explanation to the above context.

PROFILE.md
milan-zededa: Added (brief) explanation to this context (more detailed description remains below under "EVE behavior").

proto/profile/signal.proto
eriknordmark: OK

PR #145, opened 2026-05-18
config: add master_node_uuids to EdgeNodeCluster

naiming-zededa:
In a 3-master k3s cluster, replacing a master node leaves a stale
      Node object that blocks the new master from joining — k3s rejects
      a new server while an old member is still registered.

      Add master_node_uuids (field 13) to EdgeNodeCluster so the controller
      can ship the authoritative list of current master UUIDs to surviving
      devices. EVE uses this list to delete stale control-plane Node
      objects, which causes k3s to automatically drop the corresponding
      member and allows the replacement master to join cleanly.


zedi-pramodh:
I am trying to understand,  so in a replace node operation, eve will get three uuid of the two old nodes + one new node  ?
Then the eve code will delete the node that is missing from the list ? Which eve node will handle that in a cluster ? Will there be a race ?  Does this also mean our replace node never worked earlier ?

eriknordmark (approved):
LGTM.
Is this tied to using etcd as the database, or should the comment be more general?

PR #146, opened 2026-05-19
Add GPT partition identity to ZInfoStorage

eriknordmark:
## Summary

Extend `ZInfoStorage` (the per-partition entry in
`ZInfoDevice.storageList`) with three optional fields so a controller
can identify each entry by its GPT role rather than only by raw device
name:

- `partitionLabel` (5) — GPT PARTLABEL, e.g. `EFI System`, `CONFIG`,
  `IMGA`, `IMGB`, `P3`.
- `partitionTypeGuid` (6) — GPT partition-type GUID.
- `partitionUuid` (7) — GPT PARTUUID — the same identifier EVE's boot
  path uses to resolve `IMGA` / `IMGB` / `CONFIG`.

## Motivation

EVE already enumerates every block partition `lsblk` sees and reports
each one's size in `ZInfoStorage.total`. But a controller seeing
`/dev/sda1`, `/dev/sda2`, `/dev/sda3` with sizes cannot tell which is
the EFI partition, which is `IMGA` vs `IMGB`, or compare the size of
either against the rootfs being staged. With these three fields it
can.

## Compatibility

Pure additive change. New fields are appended as 5/6/7; existing
clients keep parsing the message unchanged. EVE fills the fields only
for block-partition entries on GPT-partitioned disks; directory
entries and partitions on non-GPT disks emit empty strings.

## Follow-up

EVE pillar needs a follow-up change to populate the new fields from
`lsblk -o PARTLABEL,PARTTYPE,PARTUUID` (or equivalent). That change
is not in this PR.


proto/info/info.proto
```diff
@@ -907,12 +907,35 @@ enum BaseOsSubStatus {
   UPDATE_DEFERRED = 7; // waiting for current update to finish
 }
 
-// Per filesystem/partition information
+// Per filesystem/partition information.
+//
+// Block-partition entries (device != "") carry GPT identification in
```
rene: This comment is a bit odd. Device carrying GPT identification will be block devices like `sda`, `sdb`, etc, not `sda1`, `sda2`, `sda3` which represents the partition itself.... and why `device != ""`. Please, review the comment...

eriknordmark: The device field can be either /dev/sda or /dev/sdaN and all of those are block devices in Linux. 
The comment refers to the underying disk (aka /dev/sda) being GPT. 
The GPT info is contained in /dev/sda but is associated with the partitions.

So I don't see anything wrong in the comments - maybe this can be made more clear somehow. Let me try.

eriknordmark: You're right — "block-partition" lumped whole disks and partitions together, and `device != ""` is the wrong gate (the empty-`device` case is the directory entries from `ReportDirPaths`, not whole-disk entries).

Rewrote the comments in the latest force-push: dropped the umbrella note and made each field's comment state its own emptiness rules — the GPT fields are set only when the entry is a GPT partition, and empty for whole-disk entries (e.g. `sda`), for partitions on non-GPT disks, and for directory entries.

rene (approved):

PR #147, opened 2026-06-19
config: add enable_native_k8s_orchestration to EdgeNodeCluster

naiming-zededa:
Add a boolean field to EdgeNodeCluster to enable native Kubernetes
 orchestration as an opt-in option alongside
 CLUSTER_TYPE_REPLICATED_STORAGE, replacing the role previously served
 by CLUSTER_TYPE_K3S_BASE (marked for future phase-out). Generate Go and
 Python bindings.


zedi-pramodh (commented):
Basically you are saying if REPLICATED STORAGE type just do ENC only. If REPLICATED STORAGE type plus this boolean then its ENC + ZKS combined.

Only curious question is why can't we have a separate enum for this type instead of controlling through a boolean ?

zedi-pramodh (approved):
LGTM

eriknordmark (changes requested):
@naiming-zededa can you split into two commits with the second one carrying the generated files?

cshari-zededa (approved):
LGTM, thanks @naiming-zededa

eriknordmark (approved):
LGTM

PR #148, opened 2026-06-21
Add BOOT_REASON_HW_WATCHDOG and KUBE_TRANSITION

eriknordmark:
This is moderately useful on arm64 due to the lack of support in the common Intel iTCO driver.
@rene do you know which hardware watchdog drivers are common on ARM so we can check whether the set bootreason to something other than zero?

---

Add two values to the `BootReason` enum in `info.proto`:

- **`BOOT_REASON_HW_WATCHDOG` (16)** lets a device that was reset by its
  hardware watchdog report that distinctly, instead of being reported as a
  kernel panic (`BOOT_REASON_KERNEL`) or an unknown reboot. The flag is only
  set on platforms whose watchdog driver reports `WDIOF_CARDRESET`; many
  (e.g. Intel iTCO) always report a zero boot status, so the value will not
  appear on those.
- **`BOOT_REASON_KUBE_TRANSITION` (15)** already exists in
  `pkg/pillar/types.BootReason` but was missing from this enum. Adding it
  restores the documented one-to-one correspondence with the pillar values.

Both are additive field numbers, so the change is wire-compatible — existing
controllers and existing EVE instances continue to interoperate.

The pillar code that records and consumes the new reason is in a companion
lf-edge/eve PR.

Commits follow the repo's two-commit convention: the `.proto` change is
separate from the regenerated Go/Python bindings.


rene:
@eriknordmark , SBSA (arm64 servers, NVIDIA GB10) should support the WDIOF_CARDRESET flag, but this is per driver, so we cannot assume it's always gonna be available. But in any case is good to support (check) it.

rene (approved):

PR #149, opened 2026-07-02
info: add ZDEVICE_STATE_CONVERTING state and ZDeviceSubState

eriknordmark:
## Description

Adds a device state for in-field boot-disk layout conversion (e.g. the
EVE-kvm↔EVE-k repartition), so a device repartitioning its boot disk
reports a distinct state instead of appearing generically busy or
offline.

- `ZDeviceState.ZDEVICE_STATE_CONVERTING = 9` — the device is converting
  its on-disk layout.
- `ZDeviceSubState` (new enum), reported in `ZInfoDevice.sub_state`
  (field 61) — the phase of that work. Most conversion steps run in the
  offline early-boot resizer with no controller link, so only the phases
  EVE can actually report are enumerated (reboot-to-resize, installing,
  reboot-to-target); EVE reports the furthest-reached one once it can
  talk to the controller again.

Both additions are additive (new enum value, new field number 61), so
the change is wire-compatible: existing controllers and existing EVE
instances continue to interoperate during rolling deploys.

Consumed by the EVE-side boot-disk conversion work (lf-edge/eve#6063),
which currently hand-carries this proto in its pillar vendor and will
switch to `make bump-eve-api` once this lands.

Regenerated Go and Python bindings are split into a separate commit per
the eve-api convention.


rene (approved):

rene:
@eriknordmark , please, solve the conflicts so we can merge....

christoph-zededa (approved):

PR #150, opened 2026-07-23
Add cluster-wide indication for network instance of cluster

naiming-zededa:
- It can distinguish the device related network instance or the cluster
      wide network instance, maybe needed for the native kubernetes
      applications usage


proto/config/netinst.proto
```diff
@@ -326,6 +326,19 @@ message NetworkInstanceConfig {
...
+  // If false (default), the network instance configuration is only downloaded
+  // to and instantiated on this individual device.
+  // If true, the network instance configuration is downloaded to every device
+  // that is a member of the Edge Node Cluster, and the network instance is
+  // instantiated identically on each of them. This allows application
+  // instances running on any node of the cluster to attach to the same
+  // logical network instance, which is required e.g. for app instances that
+  // may be relocated between cluster nodes. This may be useful for native
+  // kubernetes applications using the network instance.
+  bool cluster_wide = 53;
```
milan-zededa, line 341: Does `cluster_wide` mean that each device gets its own isolated replica of the network instance (with a different UUID but the same configuration)? If so, I'm not sure `cluster_wide` is the best name, since the network instance doesn't actually provide connectivity between applications running on different devices.

naiming-zededa: No. this will be no different from today's ENC, same UUID to all the nodes. But this is preparing for native kubernetes applications, which does not have any application reference to detail of NI inside pillar. This is to make a difference from device only NI, like those default NI generated.

eriknordmark (approved):
LGTM

PR #151, opened 2026-07-28
capabilities: add capability for adding/removing network interface

christoph-zededa:
Add API_CAPABILITY_APP_INSTANCE_NET_INTERFACE_CHANGE to APICapability
    
Advertises that the device can add and remove network interfaces of an
edge app instance without a purge. The app instance is restarted instead
of purged, so the controller can apply interface changes without
recreating the app's volumes.

Only merge after https://github.com/lf-edge/eve/pull/6188 has been merged


eriknordmark (approved):
LGTM

PR #152, opened 2026-08-06
attest: add ATTEST_VERSION_TYPE_EVE_INSTALL

shjala:
Reports the EVE version the device was installed with, as the basename of
the /config origin file (origin.<date>.<version>). That file is written at
install and never changed by later updates.

PCR 14 measures the /config partition, and the origin file's name is part
of that measurement, so a controller predicting PCR 14 needs the
install-time value rather than the version currently running.

Sent as another entry in ZAttestQuote.versions, so no change to the quote
message itself.


proto/attest/attest.proto
```diff
@@ -89,11 +89,13 @@ enum AttestVersionType {
    ATTEST_VERSION_TYPE_INVALID = 0;
    ATTEST_VERSION_TYPE_EVE = 1;      // BaseOS version
    ATTEST_VERSION_TYPE_FIRMWARE = 2; // Firmware version (e.g. UEFI)
+   ATTEST_VERSION_TYPE_EVE_INSTALL = 3; // EVE version at install time
 }
 
 //examples
 //ATTEST_VERSION_TYPE_EVE, "5.6.1-amd64"
 //ATTEST_VERSION_TYPE_FIRMWARE, "American Megatrends Inc.-5.12"
+//ATTEST_VERSION_TYPE_EVE_INSTALL, "origin.20240115.13.4.0"
```
eriknordmark: FWIW I get a slightly different format: /config/origin.2024-04-13.12.0.0
I don't know if we've changed the date formatting or if you have a typo.

proto/attest/attest.proto
```diff
@@ -89,11 +89,13 @@ enum AttestVersionType {
    ATTEST_VERSION_TYPE_INVALID = 0;
    ATTEST_VERSION_TYPE_EVE = 1;      // BaseOS version
    ATTEST_VERSION_TYPE_FIRMWARE = 2; // Firmware version (e.g. UEFI)
+   ATTEST_VERSION_TYPE_EVE_INSTALL = 3; // EVE version at install time
```
eriknordmark: Given that the string is not a version but a date + part of the version (no "-kvm-amd64"), maybe we should call this "eve_origin"?

proto/attest/attest.proto
shjala: fixed.

proto/attest/attest.proto
shjala: done.

eriknordmark (approved):
LGTM

